PR #9, opened 2016-05-30
fix typo

sjansen:
Found while exploring the API. It took me a few minutes to notice the field name "Schdules" contained a typo.


ranjib:
thanks @sjansen , this library is under active development, please expect changes till we cut a 0.1 release

PR #10, opened 2016-06-14
Added APIListObject to Option types to allow setting offset and

jareksm:
page through more than the inital 25 returned entries.
Also renamed a file with typo and gofmt changed ordering of some
of the imports.


ranjib:
thanks @jareksm

PR #12, opened 2016-07-13
Add team_ids[] query string arg

marklap:
I added the field in the ListUserOptions struct that produces the team_ids[] query string param. I also cleaned up the APIListObject struct adding a field tag to make the resultant query string params conform to other qs params (i.e. lowercase and omitempty).


ranjib:
thanks dude.

PR #23, opened 2016-09-03
Refactor to work correctly with V2 API

dthagard:
* Overhauled the SDK to work correctly with the PagerDuty v2 API.
* Eliminated all golint warnings including adding documentation.
* Hid some methods on Client that should not be exposed.
* Changed behavior on create/update of objects to return the resultant object.
* Fixed some struct definitions so that they are serialized to JSON correctly.
* Implemented some missing methods.
* Vendored external required libraries.


Luzifer:
> Vendored external required libraries.

Please don't do this. This will cause trouble when vendoring your library into a binary project.

> In Go, dependency management is a concern of the binary author. Libraries with vendored dependencies are very difficult to use; so difficult that it is probably better said that they are impossible to use. There are many corner cases and edge conditions that have played out in the months since vendoring was officially introduced in 1.5. (You can dig in to one of these forum posts if you’re particularly interested in the details.) Without getting too deep in the weeds, the lesson is clear: libraries should never vendor dependencies.
> (Source: [Go best practices, six years in](https://peter.bourgon.org/go-best-practices-2016/#dependency-management))

As an example: When my project uses (and vendors) `aws/aws-sdk-go` and also a library I'm using uses that SDK, we both vendor different versions this will just blow up as there are conflicting versions. Also for example if I'm using logrus (I saw you've vendored that one) Go has to decide whether to use the version in `vendor/github.com/Sirupsen/logrus` OR the version in `vendor/github.com/PagerDuty/go-pagerduty/vendor/github.com/Sirupsen/logrus`…

In the end: this will cause trouble…

ranjib:
Awesome. Dan, thanks for all the hard work. As Knut mentioned, can you
remove the vendor bits. I am planning to nuke logrus dependency. I want to
keep the library with less dependencies (I can only see go-querystring
being required as of now )

Luzifer:
Yep definitely! Thank you very much for rewriting this. As far as I've seen the code this looks very good to me. 👍

ranjib:
:trophy:

PR #27, opened 2016-09-05
Updated Readme and fixed compilation errors

seizadi:
I don't know the status of the project as it was not compiling but after I fixed the issues, at least the basic API is working. I will update the repo as I test the features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,38 +5,47 @@ go-pagerduty is a CLI and [go](https://golang.org/) client library for [PagerDut
 
 ## Installation
 
+### TO-DO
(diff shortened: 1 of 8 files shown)


ranjib:
@seizadi thank you so much for the contribution. This library under active development. I merged couple of community contributions today, which broke the code base. go-pagerduty is intended to provide a go client library to interact with PagerDuty and a CLI (that is written using the client binding).  I have cherry picked the fix. I am not merging the README related commit, as we intend this to be a client library first, and the CLI written on top of it. I'll update the README later with CLI usage.
thanks again,

PR #28, opened 2016-09-08
fix misspelling in trigger_summary_data's JSON key.

tomwans:
This is a simple misspelling that I caught while reading the documentation.


ranjib:
thank you @tomwans

PR #34, opened 2016-09-17
Fix misspelling in User struct and add JSON tags

heimweh:
Hi,
I'm using this library in a Terraform provider and noticed that I can't really create a user with the current `User` struct.

I modified the `User` struct like this to get it to work. 

**Please note that I'm very new to Go and my code might have faults in it, 
but at least I got it to work with this change.**

### DIFF
--- a/user.go
+++ b/user.go
@@ -28,7 +28,6 @@ type NotificationRule struct {
 // User is a member of a PagerDuty account that has the ability to interact with incidents and other data on the account.
 type User struct {
 	APIObject
-	ID                string `json:"id"`
 	Name              string `json:"name"`
 	Email             string `json:"email"`
 	Timezone          string `json:"timezone,omitempty"`
--- a/user.go
+++ b/user.go
@@ -28,17 +28,18 @@ type NotificationRule struct {
 // User is a member of a PagerDuty account that has the ability to interact with incidents and other data on the account.
 type User struct {
 	APIObject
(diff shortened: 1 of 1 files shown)


heimweh:
The original error I got from the API without the modified struct.

```
$ terraform apply
pagerduty_user.foo: Creating...
  email:     "" => "foo@bar.com"
  job_title: "" => "Job title"
  name:      "" => "Foo Bar"
Error applying plan:

1 error(s) occurred:

* pagerduty_user.foo: Failed call API endpoint. HTTP response code: 400. Error: &{2001 Invalid Input Provided [Name must be a String.]}
```

user.go
```diff
@@ -28,17 +28,18 @@ type NotificationRule struct {
...
 type User struct {
 	APIObject
-	Name              string
-	Email             string
-	Timezone          string
-	Color             string
-	Role              string
-	AvatarURL         string `json:"avatar_url"`
-	Description       string
+	ID                string `json:"id"`
```
ranjib: @heimweh ID comes from APIObject struct, which we are embedding here. You can omit this attribute

heimweh:
@ranjib thanks for the response, I'll have a look at getting this fixed asap.

heimweh:
@ranjib should be fixed now

ranjib:
thank you :-)

PR #37, opened 2016-09-24
Skip JSON encoding if no payload was given

heimweh:
This solves the issue when calling `AddUserToTeam()`, 
but there might be a better/cleaner way of solving this.

The API response (if successful returns 204) and no payload is necessary when calling `PUT	/teams/{id}/users/{user_id}`.

https://v2.developer.pagerduty.com/v2/page/api-reference#!/Teams/put_teams_id_users_user_id

### DIFF
--- a/client.go
+++ b/client.go
@@ -59,11 +59,15 @@ func (c *Client) delete(path string) (*http.Response, error) {
 }
 
 func (c *Client) put(path string, payload interface{}, headers *map[string]string) (*http.Response, error) {
-	data, err := json.Marshal(payload)
-	if err != nil {
-		return nil, err
+
+	if payload != nil {
+		data, err := json.Marshal(payload)
+		if err != nil {
+			return nil, err
+		}
+		return c.do("PUT", path, bytes.NewBuffer(data), headers)
 	}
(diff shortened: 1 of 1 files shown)


heimweh:
Also, without this change I get back this:
```go
* pagerduty_user.foo: Response did not contain formatted error: Could not decode JSON response: json: cannot unmarshal string into Go value of type pagerduty.errorObject. HTTP response code: 400. Raw response: &{Status:400 Bad Request StatusCode:400 Proto:HTTP/1.1 ProtoMajor:1 ProtoMinor:1 Header:map[Server:[nginx] Content-Type:[application/json; charset=utf-8] Date:[Sat, 24 Sep 2016 07:45:17 GMT] Connection:[keep-alive] Status:[400 Bad Request] X-Ua-Compatible:[IE=Edge,chrome=1] Cache-Control:[no-cache]] Body:0xc420403900 ContentLength:-1 TransferEncoding:[chunked] Close:false Uncompressed:false Trailer:map[] Request:0xc42038c3c0 TLS:0xc4200aa210}
```

ranjib:
thanks @heimweh

PR #38, opened 2016-09-28
Remove subdomain argument from escalation_policy example.

cmluciano:
PR #19 removed the subdomain from pagerduty.NewClient. This example was
not changed and fails the build when running the build task of the
Makefile.


ranjib:
thanks @cmluciano

PR #48, opened 2016-10-19
Augment Channel for other possible fields

jakedavis:
Hi!

We've found ourselves really wanting the "Details" from the channel field within the first trigger log entry, and noticed that the struct in this library doesn't include it as a field. I'm hoping we can augment that struct so we can access it within our project. The commit contains a bit more context as well.

I couldn't find much documentation about channels, so I don't know if this will cause any bad effects, but it _appears_ to me that these four fields are consistent.

Apologies for the little whitespace change; I can recommit without if need be. My go linting tool does that automatically.

Thanks for reviewing!

### DIFF
--- a/log_entry.go
+++ b/log_entry.go
@@ -2,6 +2,7 @@ package pagerduty
 
 import (
 	"fmt"
+
 	"github.com/google/go-querystring/query"
 )
 
@@ -10,7 +11,10 @@ type Agent APIObject
 
 // Channel is the means by which the action was carried out.
 type Channel struct {
-	Type string
+	Type    string
+	Details string
+	Subject string
+	Summary string
 }
 
 // LogEntry is a list of all of the events that happened to an incident.


jakedavis:
I managed to find some more information from the v1 API documentation, and it looks like I've configured it specifically for the `Web Trigger` type:

https://v1.developer.pagerduty.com/documentation/rest/log_entries/show

So it sounds like this PR won't work as-is, but maybe I can instead change Channel to be a `*json.RawMessage` instead?

ranjib:
@jakedavis that is exactly what i am thinking. We would be loosing the known "type" entry in this route, i tried this without success
```go
type Channel struct{
json.RawMessage
 Type string `json:"type"`
}
```

stmcallister:
We completely agree this should be addressed, but as mentioned above not with this implementation. Going to close this now, but would welcome additional ideas going forward.

PR #49, opened 2016-10-20
Add vendor support

heimweh:
Hi, this PR adds support for Vendor (`list` & `get`). 
The listing of vendors seems to work fine, but I have not yet tried the get function.

https://v2.developer.pagerduty.com/v2/page/api-reference#!/Vendors/get_vendors

### DIFF
--- a/vendor.go
+++ b/vendor.go
@@ -1,0 +1,73 @@
+package pagerduty
+
+import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


ranjib:
thank you.

ranjib:
thank you.

PR #52, opened 2016-11-12
Schedule restriction - Add support for start_day_of_week

heimweh:
Hi,

this adds support for `start_day_of_week`. 
Without `start_day_of_week` set, a restriction with type `weekly_restriction` will fail with:

```
Failed call API endpoint. HTTP response code: 400. Error: &{2001 Invalid Input Provided [Start day of week must be in 1, 2, 3, 4, 5, 6, 7, 1, 2, 3, 4, 5, 6, or 7.]}
```

### DIFF
--- a/schedule.go
+++ b/schedule.go
@@ -2,14 +2,16 @@ package pagerduty
 
 import (
 	"fmt"
-	"github.com/google/go-querystring/query"
 	"net/http"
+
+	"github.com/google/go-querystring/query"
 )
 
 // Restriction limits on-call responsibility for a layer to certain times of the day or week.
 type Restriction struct {
 	Type            string `json:"type,omitempty"`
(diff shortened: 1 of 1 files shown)


heimweh:
Hi @ranjib. Would you be able to take a look at this one? 🙂 Not sure if I missed something.

blalor:
Bump; would love to see this make it into Terraform (and made available to other users of the library).

fmasuhr:
@ranjib I would also like to see this merged to use this via terraform :)

paultyng:
Interestingly the field is missing in the online docs as well: 

https://v2.developer.pagerduty.com/v2/page/api-reference#!/Schedules/get_schedules

evan2645:
Thank you for this, and very sorry for the delay.

Also thanks for pointing out that this option is missing in the docs @paultyng we have filed an internal ticket to get that fixed up

PR #53, opened 2016-11-17
Allow service timeouts to be disabled

heimweh:
This should allow both `AutoResolveTimeout` and `AcknowledgementTimeout` to be disabled. 

However this does break the ability to just use the API defaults (which is enabled).
Maybe instead of this, we could have an enabled/disabled bool flag?

### DIFF
--- a/service.go
+++ b/service.go
@@ -2,8 +2,9 @@ package pagerduty
 
 import (
 	"fmt"
-	"github.com/google/go-querystring/query"
 	"net/http"
+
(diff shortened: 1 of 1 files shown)


heimweh:
Thanks @jtsaito for looking into this, almost forgot about this one. 
Would be awesome to get this merged. 

@ranjib any chance you could take a look at this one? :) many thanks

jtsaito:
@heimweh @brhou Thank you very much for looking into this!

PR #59, opened 2016-12-20
Added missing slash to delete integration method url

jescochu:
Seems the `DeleteIntegration` method gives a 404 due to a missing slash before the integration ID query param. This patches that up.


ranjib:
thanks @jescochu

PR #60, opened 2017-01-16
fix since and until by adding them to url scheme

ethansommer:
The ListLogEntriesOptions didn't have url variable names defined for since or until, and thus ignored those options. (well, it actually put omitempty=TIME, but that didn't do anything)

This change appears to fix that.


gkuchta:
👍

beono:
@ranjib Could you please merge it?

PR #62, opened 2017-01-27
Explicitely set service attributes to null

jtsaito:
This PR addresses a change in the behaviour `/services` `POST` endpoint in the V2 API of Pagerduty (see [documentation](https://v2.developer.pagerduty.com/v2/page/api-reference#!/Services/post_services)). Due to the change we had issues with our Terraform setup.

**Problem**
When creating a service without `acknowledgement_timeout` or `auto_resolve_timeout`, the service has acknowledgement timeout feature or auto resolve timeout feature enabled respectively. This was not the case so far. There is no longer a way to create a resource with the features disabled.

Expected behaviour:
When creating a `pagerduty.Service` without `acknowledgement_timeout` or `auto_resolve_timeout` by calling `client.CreateService(service)`, the service disables the features and does not set `acknowledgement_timeout` and `auto_resolve_timeout`. (This used to be the case until a few days ago.)

Observed behaviour:
The service enables the features and sets non-zero default values for `acknowledgement_timeout` and `auto_resolve_timeout`.

**Example code for reproducing the issue**
```
package main

import (
  "fmt"

  pagerduty "github.com/PagerDuty/go-pagerduty"
)

func main() {
  client := pagerduty.NewClient("YOUR-API-TOKEN-HERE")

  epo := pagerduty.GetEscalationPolicyOptions{}
  ep, epErr := client.GetEscalationPolicy("PH77HFE", &epo)

  if epErr != nil {
    fmt.Printf("error: %v\n", epErr)
  }

  s := pagerduty.Service{
    Name:               "Cha0tic App",
    Description:        "Some infinities are bigger than other infinities",
    EscalationPolicy:   *ep,
    AutoResolveTimeout: nil,
  }

  t, e := client.CreateService(s)
  if e != nil {
    fmt.Printf("error: %v\n", e)
  }

  fmt.Println(t)
}
```

**Submitted solution**
We drop `omitempty` for the attributes in question. This restores the default behaviour of `client.CreateService(service)`.

### DIFF
--- a/service.go
+++ b/service.go
@@ -2,8 +2,9 @@ package pagerduty
 
 import (
 	"fmt"
-	"github.com/google/go-querystring/query"
 	"net/http"
+
+	"github.com/google/go-querystring/query"
 )
 
 // Integration is an endpoint (like Nagios, email, or an API call) that generates events, which are normalized and de-duplicated by PagerDuty to create incidents.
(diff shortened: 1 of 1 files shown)


jtsaito:
Sorry, there has already been a PR for the same issue: https://github.com/PagerDuty/go-pagerduty/pull/53. Please merge that.

Closing this one.

PR #64, opened 2017-02-02
Add query to vendor

heimweh:
This PR adds support for querying vendors.

This allows us to easier find vendors by name.

Example:
```go
client.ListVendors(&pagerduty.ListVendorOptions{
    Query: "Datadog",
}
```

Tested and seems to work quite well.


heimweh:
Would be awesome to be able to query vendors by name like this. 
Would any of you mind taking a look? :)

I'm not 100% sure if there are other options that should be in there as well.

//cc @ranjib @Deconstrained

PR #65, opened 2017-02-06
Update event.go - fix a memory leak

AngelRefael:
To prevent a memory leak the response body must be closed even when it is not used


lfepp:
Additional details on this issue can be found at: http://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body-in-golang

PR #68, opened 2017-03-02
Fixing Maintenance typo

domudall:
Was previously `CreateMaintaienanceWindows`, changed to `CreateMaintenanceWindows`, however this is a breaking API change if users are accessing the typo version.

### DIFF
--- a/command/maintenance_window_create.go
+++ b/command/maintenance_window_create.go
@@ -59,7 +59,7 @@ func (c *MaintenanceWindowCreate) Run(args []string) int {
 		return -1
 	}
 	log.Debugf("%#v", m)
-	if _, err := client.CreateMaintaienanceWindows(m); err != nil {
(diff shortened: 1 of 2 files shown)


domudall:
Also turns out that the URLs were returning 404 for a number of maintenance window functions as well:

* Create
* Delete
* Get

All had `mainteance_windows`, and have now been corrected to `maintenance_windows`.

augier:
Cor, this looks useful.

ethanfrogers:
What do we need to do to get this merged? I'm running into the same issue. Perhaps we could disregard the method signature changes and just update the endpoint to prevent breaking changes?  I just submitted a PR fixing the same thing, but then found this issue.

lfepp:
Thanks for the contribution here and it LGTM!

After taking a look at the code I don't think there will actually be any breaking changes to the library as the methods that were updated were already broken. Truly functioning client code using this library should not be affected.

PR #69, opened 2017-03-02
Fix maintenance window JSON

domudall:
Ensuring JSON fields on maintenance windows structs are explicitly set, as capitalised JSON fields are not recognised by the v2 API. Code formatted with gofmt.


lfepp:
Thanks for the contribution

PR #85, opened 2017-06-05
[incidents] fix entries typo in a few places

joeyparsons:
There are a few places where `entries` is spelled `entires` in various incidents functions.  This simply fixes those issues.


esigler:
Thanks!

PR #93, opened 2017-10-10
Added json fields to structs

bradleyrobinson:
Because some of the structs lacked json fields, some methods did not work properly with the API. For example, the pagerduty.Client.ManageIncidents() returns a 2001 error with the message: "Assignee cannot be empty" even when the Assignee field is specified. By specifying the json field, assignee is mapped with the correct name in the request and the request goes through.


wdhnl:
@nuptse can you look at this PR?

abtris:
Is some progress here? I hope, this can fix problem with Incident includes issue.

PR #96, opened 2017-10-21
List all log entries for an incident id

agriel:
This change allows you to request a list of log entries for a given incident id. This already exists in the REST API, but seems to be missing here.


agriel:
Oops. I was using an outdated version.

PR #98, opened 2017-11-26
Add oncall details

luqasn:
When you request additional details by setting `Includes` to `users,schedules`, you won't get access to the requested information because their mapped types are set to the basic `APIObject` and thus does not contain any data beyond an ID and a Summary, although the JSON response contains more details.

### DIFF
--- a/event_v2.go
+++ b/event_v2.go
@@ -1,0 +1,63 @@
+package pagerduty
+
+import (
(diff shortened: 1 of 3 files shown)


luqasn:
Closing this because I accidentally referenced the fork's master. New PR is #104

Issue #83, opened 2017-05-12
Support V2 events


tmccleve:
In order to leverage the new severity feature for PagerDuty events, this must be updated to utilize the v2 of the events api. I'm willing to do most of the work on updating. I'm looking for some guidance on how the maintainers would prefer to approach backward compatibility with the V1 events.

tmccleve:
Spiked this out for anyone that's interested. https://github.com/TWExchangeSolutions/prtg-pagerduty/tree/master/event

stmcallister:
Cleaning out old issues. This one appears to be addressed. Please repost if not.

PR #103, opened 2018-01-04
Add support for v2 ManageEvent api call

luqasn:
I took the liberty of adding the spike by @tmccleve from #83 to my fork in order to support creation of v2 API events. The code is almost identically to the v1 code that is already in the package apart from the JSON structures and the endpoint url.

### DIFF
--- a/event_v2.go
+++ b/event_v2.go
@@ -46,7 +46,7 @@ func ManageEvent(e V2Event) (*V2EventResponse, error) {
 	if err != nil {
 		return nil, err
 	}
-	req, _ := http.NewRequest("POST", eventEndPoint, bytes.NewBuffer(data))
+	req, _ := http.NewRequest("POST", v2eventEndPoint, bytes.NewBuffer(data))
 	req.Header.Set("Content-Type", "application/json")
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
--- a/event_v2.go
+++ b/event_v2.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 )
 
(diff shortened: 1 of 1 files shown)


luqasn:
Can we get this merged?

v-antech:
I would also like see if this can be merged. Looks fine to me. Also why is the library stating that it is supporting the V2 API, when in fact it is not? This seems a very essential feature to me.

dan-force:
My team could really benefit from this change being merged. How can we help to make this happen? Who is in charge, btw? Can you help us out @istvandn or @luqasn?

luqasn:
I have no idea who is in charge, but this library sadly seems to be pretty much abandoned by pagerduty. I think a new project should be created that is based on the swagger api definition for the v2 api. This would make keeping up to date much easier.

event_v2.go
```diff
@@ -0,0 +1,68 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+)
+
+// Event includes the incident/alert details
+type V2Event struct {
+	RoutingKey string        `json:"routing_key"`
+	Action     string        `json:"event_action"`
```
nuptse: Please rename this to `EventAction` for consistency with the JSON field name

event_v2.go
```diff
@@ -0,0 +1,68 @@
...
+	Severity  string      `json:"severity"`
+	Timestamp string      `json:"timestamp,omitempty"`
+	Component string      `json:"component,omitempty"`
+	Group     string      `json:"group,omitempty"`
+	Class     string      `json:"class,omitempty"`
+	Details   interface{} `json:"custom_details,omitempty"`
+}
+
+// Response is the json response body for an event
+type V2EventResponse struct {
```
nuptse: Based on our [current API documentation for the Events API v2](https://v2.developer.pagerduty.com/docs/events-api-v2), the response format is different than what is expected here.

event_v2.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, err
+	}
+	req, _ := http.NewRequest("POST", v2eventEndPoint, bytes.NewBuffer(data))
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusAccepted {
```
nuptse: The API also returns validation errors back, it'll be great to have those displayed back to the user.

mattstratton:
I haven't dug into this PR enough yet, but is it backwards compatible with v1? If not, we will want to be very clear with this release that it contains breaking changes. 

I am intending to add binary release options via goreleaser shortly; we should wait to merge this until we have the major versions cut, which I hope to do within the next couple days.

jamiehannaford:
@mattstratton Hey! Would it be possible to add this soon? Since the struct names contain `v2` I think it's backwards compatible.

PR #105, opened 2018-01-04
Add support for incident creation

luqasn:
The current version of the client does not allow creation of incidents for services. This PR changes that.

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -1,6 +1,8 @@
 package pagerduty
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 
 	"github.com/google/go-querystring/query"
@@ -41,6 +43,12 @@ type Incident struct {
 	Teams                []APIObject       `json:"teams,omitempty"`
 	Urgency              string            `json:"urgency,omitempty"`
 	Status               string            `json:"status,omitempty"`
+}
+
+type CreateIncidentOptions struct {
+	APIObject
+	Title   string       `json:"title,omitempty"`
+	Service APIReference `json:"service,omitempty"`
 }
 
 // ListIncidentsResponse is the response structure when calling the ListIncident API endpoint.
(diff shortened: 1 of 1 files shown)


sevagh:
Wouldn't the body be better as an `Incident` object? https://github.com/PagerDuty/go-pagerduty/compare/master...sevagh:master

I feel like that matches the API more: https://v2.developer.pagerduty.com/docs/incident-creation-api

Maybe I've misunderstood.

luqasn:
I don’t understand what you mean, your and my changes look almost identical to me (though I only implemented support for title)

sevagh:
Sorry about that, I did some force-pushes between my first message and now.

I added this: `EscalationPolicy APIReference `json:"escalation_policy,omitempty"`` so that I can alert on different escalation policies. Is there a way to do that from your change?

sevagh:
This is how I use it:

```
createIncidentOptions := pagerduty.CreateIncidentOptions{
                Title: "oncall-bot-incident",
                Service: pagerduty.APIReference{
                        ID:   "XXXXXX",
                        Type: "service_reference",
                },
                EscalationPolicy: pagerduty.APIReference{
                        ID:   escalation.ID,
                        Type: "escalation_policy_reference",
                },
        }
```

sevagh:
What do you think of adding the `EscalationPolicy` field?

```
+type CreateIncidentOptions struct {
+	APIObject
+	Title    string   `json:"title,omitempty"`
+        Service APIReference `json:"service,omitempty"`
+        EscalationPolicy APIReference `json:"escalation_policy,omitempty"`
+}
+```

This is how I'm using it in prod right now. https://v2.developer.pagerduty.com/docs/incident-creation-api#making-a-request

`EscalationPolicy` is one of the optional fields in the API spec. What do you think about adding it to your PR? Then I don't need to use my fork when your PR gets merged.

luqasn:
Yeah, I think that's a good idea. Will try to get to that in the next days.

luqasn:
I have added most of the other supported parameters, but a few are still missing, for example a detailed message body. This is good enough for my use case, but I think for completeness, I should add everything else as well.

eldios:
I guess that this can be closed since the function has been already added, right @mattstratton @luqasn ?

luqasn:
Yep, closing. Thanks for adding it.

Issue #110, opened 2018-02-17
Custom connection settings


bwester:
Currently, the `Client` in this package only allows connections to PagerDuty's live service using the default Go connection settings. It would be extremely useful to me to be able to create a client with custom connection settings. Here are two specific use cases I had in mind:

1) I would like to be able to use this package in a service that has to go through an HTTP proxy to get to the outside Internet. The standard `net/http` package lets me set up connections through a proxy via [`Transport.Proxy`](https://golang.org/pkg/net/http/#Transport). But to make use of this setting, I need a `pagerduty.Client` that can use an arbitrary `*http.Client` instead of the default.

2) I would like to be able to test my code that uses a `Client` without actually operating on my real PagerDuty account. One way to achieve this goal is to write HTTP server stubs and have the client contact my testing server (e.g., from `net/http/httptest`). To do that, I need a client that can send its requests to an arbitrary endpoint, not always `api.pagerduty.com`.

Are there any existing plans that address these needs? I'd be happy to send in a pull request if needed.

rapenchukd:
I actually ran into this today, I'm including the event function into an app I am writing, and noticed there is no good way to set an http proxy that isn't global. the easiest way to do this would be to replace instances of `http.DefaultClient.Do` with `http.Client.Do` and allow users to override client/transport options in `Client{}`

I am working on a PR to do this now.

mattstratton:
I’m just looking at writing tests for this project, so use case #2 is key. I’ll review the existing PR and see if it looks mergable as-is.

theckman:
For #2, I wonder if using the PagerDuty client for this is not the best path. I think a better pattern would be to have your code accept an interface that matches the functionality exposed by the PagerDuty API client, you can then write your own mock implementation(s) that returns the data you want. That way instead of testing the PagerDuty API client, on top of your code, you're only testing your code.

theckman:
So I've raised a PR (#121) which should address usecase 1 in this issue. It's similar to PR #112, except it allows consumers to use more than `*http.Client` by accepting the client as an interface.

I think as a project we should determine whether usecsase 2 is something we want to support. I'm of the personal opinion that it isn't something we should support, and should instead encourage consumers to use dependency injection via interfaces. This should allow consumers to mock the client's behavior while unit testing.

bwester:
Thank you @theckman! Your PR looks great to me, and I agree that unit testing via interfaces is probably a better solution for (2). 👍

PR #112, opened 2018-03-02
Allow caller to specify a *http.Client

markjen:
Allow *http.Client to be injected so that go-pagerduty can be used in environments where the http.DefaultClient cannot be used (e.g. Google AppEngine standard environment)

### DIFF
--- a/client.go
+++ b/client.go
@@ -45,12 +45,19 @@ type errorObject struct {
 // Client wraps http client
 type Client struct {
 	authToken string
+	client    *http.Client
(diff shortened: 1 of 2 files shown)


rapenchukd:
Big +1 here, running into concerns using this in production for various reasons. being able to configure custom client is absolutely needed.

Just commented on #110 looking for this, you saved me having to write some code :)

mattstratton:
Closing this in favor of #121 ...

PR #116, opened 2018-05-01
On call return user

LucaPericoIT:
on-call call does return a simple _ApiObject_ as User. I changed it for an _User_ object


LucaPericoIT:
API does not return the User object

PR #119, opened 2018-05-30
Add CreateIncident to API and cli

wdhnl:
There is error checking in CreateIncident that more aggressive than other functions, not sure if that should be removed for consistency.

CreateIncidentResponse is a new exported struct, but in CreateIncident, this code returns (*Incident, error), not sure if should make it non-exported, or have CreateIncident return it (although the client will only care about the Incident part of the response)


wdhnl:
closing this because I think https://github.com/PagerDuty/go-pagerduty/pull/122 is more comprehensive

PR #120, opened 2018-06-01
extract root files into package

buddhamagnet:
First step in restructure, extract root files into ```pagerduty``` folder.

### DIFF
--- a/examples/escalation_policies.go
+++ b/examples/escalation_policies.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"fmt"
-	"github.com/PagerDuty/go-pagerduty"
+	"github.com/PagerDuty/go-pagerduty/pagerduty"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


rogerhub:
Er... why? Is there any value in this that justifies breaking all callers?

theckman:
Hm. Yeah. What value are we looking to add with this change? I am a proponent of using sub-packages when they make sense, I don't know if it applies here.

theckman:
@buddhamagnet To someone like @rogerhub, this PR makes a major breaking change without any context justifying the change within the commit message. This reads like a direct question, trying to get some context (justification) behind this change.

This may be a culture thing, but I personally didn't perceive impoliteness in that comment. Maybe that is because I am more direct in my communication.

I'm sorry if the way the questions were asked around the change made you feel like closing it outright. If you'd like to raise any PRs in the future we would be happy to review them.

PR #121, opened 2018-06-01
Allow package consumers to provide their own HTTP client

### LINKED ISSUE #110: Custom connection settings

theckman:
This updates the package to no longer make use of the `http.DefaultClient`, and
to instead use a package-local default client as well as to provide consumers
the ability to provide their own client in the time of need.

The reason for moving away from the `http.DefaultClient` is for two reasons:

1. It prevents other packages from being able to impact this one by making
   changes to the `http.DefaultClient`.

2. The isolation should help make this package more resilient, as we can tweak
   the client's transport settings to provide defaults appropriate for PagerDuty
   usage.

The package-local default client is not directly exposed, so as to hopefully
avoid people making out-of-band changes to its config and to instead provide
their own client.

This also supports the ability for consumers to provide their own HTTP client,
in the situation where the default client doesn't have the settings they need.
The ability to provide this client was done using an interface, `HTTPClient`, so
that consumers aren't forced to make use of `*http.Client` if they don't want
to.

Fixes #110

Signed-off-by: Tim Heckman <t@heckman.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,11 @@ func main() {
 }
 ```
 
+The PagerDuty API client also exposes its HTTP client as the `HTTPClient` field.
+If you need to use your own HTTP client, for doing things like defining your own
+transport settings, you can replace the default HTTP client with your own by
+simply by setting a new value in the `HTTPClient` field.
+
 ## License
 [Apache 2](http://www.apache.org/licenses/LICENSE-2.0)
 
--- a/client.go
+++ b/client.go
@@ -5,7 +5,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"runtime"
+	"time"
 )
 
 const (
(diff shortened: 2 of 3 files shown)


theckman:
This PR is similar to #112. The difference here is that instead of taking an `*http.Client`, it takes an interface value that implements the `Do` method signature of an `*http.Client` (`Do(*http.Request) (*http.Response, error)`). This opens us up to support more than just `*http.Client`.

mattstratton:
👍

PR #127, opened 2018-07-20
CreateIncident takes CreateIncidentOptions param

wdhnl:


eldios:
@mattstratton could you please go ahead and approve this PR and release it?

It's just a quick typo-fix.

Thanks.

asf-stripe:
I don't think that fixes anything? According to the API docs, you have to wrap the `CreateIncidentOptions` in the `CreateIncident` struct so that the correct object gets POSTed... we're using the original form of this and incidents are correctly getting generated.

PR #133, opened 2018-11-08
fix the user time_zone

chunyanlv:
according to https://v2.developer.pagerduty.com/v2/page/api-reference#!/Users/get_users_id, it is "time_zone" not "timezone"


stmcallister:
This appears to have been fixed by another PR. Going to close.

PR #140, opened 2019-03-13
Remove useless else from README

soullivaneuh:
As the panic function will kill the program, the else statement is useless.

Am I right? :-)


soullivaneuh:
Well, it was not really true as the variable is declared on the if scope.

Here is an another proposition I found more clear for documentation. What do you think?

PR #143, opened 2019-03-14
Incident alerts

soullivaneuh:
Continue of #142

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -176,27 +176,33 @@ func (c *Client) ListIncidentNotes(id string) ([]IncidentNote, error) {
 
 // IncidentAlert is a alert for the specified incident.
 type IncidentAlert struct {
-	ID        string    `json:"id,omitempty"`
-	Summary	  string    `json:"summary,omitempty"`
-	CreatedAt string    `json:"created_at,omitempty"`
-	AlertKey  string    `json:"alert_key,omitempty"`
+	APIObject
+	CreatedAt   string                 `json:"created_at,omitempty"`
+	Status      string                 `json:"status,omitempty"`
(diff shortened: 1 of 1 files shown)


incident.go
```diff
@@ -176,15 +176,21 @@ func (c *Client) ListIncidentNotes(id string) ([]IncidentNote, error) {
...
-	ID        string    `json:"id,omitempty"`
-	Summary	  string    `json:"summary,omitempty"`
-	CreatedAt string    `json:"created_at,omitempty"`
-	AlertKey  string    `json:"alert_key,omitempty"`
+	APIObject
+	CreatedAt   string       `json:"created_at,omitempty"`
+	Status      string       `json:"status,omitempty"`
+	AlertKey    string       `json:"alert_key,omitempty"`
+	Service     APIObject    `json:"service,omitempty"`
+	Body        interface{}  `json:"body,omitempty"`
```
soullivaneuh: I don't know how to deal with that not working field.

It has various of data even a key that are not represented on the official API documentation like `cef_details`.

I would like to store it as a "Various mixed array", but I don't know how to do to make json unmarshal happy.

Tried `interface{}` and `[]interface{}` without success:

```
json: cannot unmarshal number into Go value of type []pagerduty.IncidentAlert
```

Any help would be appreciated.

soullivaneuh:
I finally found the issue: I simply forgot to manage the pagination.

mattstratton:
Can you please rebase this so that the CircleCI config gets into your branch so that the Circle tasks will run?

soullivaneuh:
@mattstratton Done.

The only missing thing is a pagination, but I don't really know how to handle it.

Shall I create a new `ListIncidentAlertsOptions`?

PR #145, opened 2019-03-21
Add optional summary field for log entry channel

soullivaneuh:
The field appears on some channel type like `note` for example.


stmcallister:
Thanks for submitting this, and sorry for the delay in responding! I went with a different approach in #207 .  But, this accomplished the same end.

PR #147, opened 2019-04-21
Add the ability to set urgency via the CreateIncident api call

baveltman:
This PR adds the urgency field to CreateIncidentOptions struct, such that users can now specify urgency of "low" or "high" when creating an incident via the CreateIncident api.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
 bin/*
 *.swp
 .idea
-go.mod
-go.sum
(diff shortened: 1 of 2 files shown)


stmcallister:
Thank you for making this contribution! However, i just merged [PR 151](https://github.com/PagerDuty/go-pagerduty/pull/151) which also added this field as well as a few others.

PR #152, opened 2019-05-31
Work around the LogEntry.Channel field being not well defined.

seanhoughton:
In order to get access to the event details you need to look into the contents of the channel field. The current go implementation loses all that data because it's "Channel" type only includes the "Type" field. By making this an interface we can walk the dictionary and access the details. I'm not sure if this is the best solution, but I feel it's better than preventing access entirely...

Note: APIs that return different types in the same field make working with them in statically typed languages painful. Maybe you can provide some feedback to the API developers while they are still undecided on the structure of this field.

### DIFF
--- a/log_entry.go
+++ b/log_entry.go
@@ -10,9 +10,14 @@ import (
 type Agent APIObject
 
 // Channel is the means by which the action was carried out.
-type Channel struct {
-	Type string
-}
+//type Channel struct {
+//	Type string
+//}
+
+// Channel is not well defined and uses a variadic type.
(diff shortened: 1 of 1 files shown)


stmcallister:
Sorry for the delay! Thanks for submitting this PR! I like your approach, except for that it breaks the Channel object for users currently using it. Instead, I [added an additional field](https://github.com/PagerDuty/go-pagerduty/pull/207) to Channel called Raw and have that be the `map[string]interface{}`. That way we keep everything working, and add the ability to access the varying fields of the `Channel` object.

PR #154, opened 2019-06-19
Support Links in V2Event

alindeman:
https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2#section-the-links-property


stmcallister:
Thank you for adding Links to the Event struct!

PR #157, opened 2019-07-16
omit incident fields in payload

rizzza:
Need to use struct pointers to omit these fields when creating an incident.

```
curl -X POST -H "From: test@someone.com" -H 'Accept: application/vnd.pagerduty+json;version=2' -H 'Authorization: Token token=<secret>' -H "content-type: application/json" https://api.pagerduty.com/incidents -d '{"incident":{"type":"incident","title":"Help me obiwan","service":{"id":"<id>","type":"service_reference"},"priority":{},"incident_key":"","body":{},"escalation_policy":{"id":"<id>","type":"escalation_policy_reference"}}}'

{"error":{"message":"Invalid Input Provided","code":2001,"errors":["Body is invalid."]}}%
```

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -95,13 +95,13 @@ type createIncidentResponse struct {
 
 // CreateIncidentOptions is the structure used when POSTing to the CreateIncident API endpoint.
 type CreateIncidentOptions struct {
-	Type             string       `json:"type"`
-	Title            string       `json:"title"`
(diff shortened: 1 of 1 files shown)


stmcallister:
Thanks for submitting this fix! However, this same fix was already done in a previously merged PR.

Issue #135, opened 2019-02-14
Priority field should be optional according to API spec


auhlig:
Managing an incident should not require the `priority` field according to the [API spec](https://api-reference.pagerduty.com/#!/Incidents/put_incidents). It's marked as optional.
However, trying to manage an incident without this field returns:
```
err="Failed call API endpoint. HTTP response code: 400. Error: &{2001 Invalid Input Provided [Priority id cannot be empty.]}"
```
Moreover, the incident priority feature is disabled in our Pagerduty.
Can you confirm that's a bug?  Kindly asking for help here @mattstratton 🙏

atomicules:
I don't think this is a bug in the API, but a bug introduced in this go-pagerduty library. I say that because it's possible to roll back and use an earlier version of this library and have things work.

I have not tracked down exactly what is causing it, but I suspect related to this commit:

- https://github.com/PagerDuty/go-pagerduty/commit/c94ddd7ce3165dc07107d141134363d6e6b67eb1

So... via Go modules I can have this in `go.mod` to use an earlier version:

```go
module stuff
require (
        github.com/google/go-querystring v1.0.0
        github.com/pagerduty/go-pagerduty v0.0.0-20180528123509-b4a4067bdbde
)
```

And then (difficult to communicate this via snippets of code, but...) I have some code which we use for batch resolving alerts:

```go
type incident struct {
	id, summary string
}
[...]
incidents := make([]incident, 0)
[...]
inctoresolve := pagerduty.Incident{
	APIObject: pagerduty.APIObject{
		ID:   incidents[i].id,
		Type: "incident_reference",
	},
	Status: "resolved",
}
[...]
batchIncs = append(batchIncs, incstoresolve)
for b := range batchIncs {
	if err := client.ManageIncidents(*email, batchIncs[b]); err != nil {
		log.Println(err)
	} else {
		log.Println("Resolving incidents...")
	}
}
```

And this works fine.

But if I use a recent version of this library:

```
module stuff

require (
        github.com/google/go-querystring v1.0.0
        github.com/pagerduty/go-pagerduty v0.0.0-20190820225956-768e5bce92a6
)
```

then I hit the same `[Priority id cannot be empty.]` issue. If I tweak the code so it's like this:

```go
inctoresolve := pagerduty.Incident{
	APIObject: pagerduty.APIObject{
		ID:   incidents[i].id,
		Type: "incident_reference",
	},
	Priority: pagerduty.Priority{
[...]

PR #164, opened 2019-08-21
Use pointer to Priority so don't send an empty priority for incident

atomicules:
Due to the changes to the Incident struct the library now inadvertently
sends an empty priority to the API (`"priority":{}`) and for
ManageIncidents this results in a response of:

    Priority id cannot be empty

To prevent sending an empty priority we can use a pointer to Priority.

Seems like a super simple fix for this - and as such I'm worried I've
missed something else.

Full discussion of this in this issue:

References: https://github.com/PagerDuty/go-pagerduty/issues/135


auhlig:
🎉

stmcallister:
thank you for contributing this change!

Issue #165, opened 2019-08-27
failed to create an incident


chunyanlv:
Here is my code:
```
        opts := officialPD.CreateIncidentOptions{
                Type: "incident",
                Title: title,
                Service: &service,
                Body: &body,
                EscalationPolicy: &escalationPolicy,
        }
 _, err := client.CreateIncident(INCIDENT_FROM, &opts)
```
I have the title correctly set.
But I got the error like:
```
panic: Failed call API endpoint. HTTP response code: 400. Error: &{2001 Invalid Input Provided [Title cannot be empty.]}
```
Does anyone know what is wrong?

snaveen:
I saw your fix. It would be really helpful for me if this gets merged. Is there any workaround till that?

joepurdy:
@snaveen My PR being merged is up to the maintainers of PagerDuty/go-pagerduty

If you want to use my fork for the time being I've gone and merged my fix into master there: https://github.com/joepurdy/go-pagerduty

I'll do my best to keep the fork updated with upstream, but ultimately the best fix would be the maintainers merging the fix into upstream.

snaveen:
Thanks @joepurdy

theckman:
I think I have merge access, let me see if I can do the needful tonight (Pacific time).

snaveen:
> I think I have merge access, let me see if I can do the needful tonight (Pacific time).

@theckman could you help us on this?

snaveen:
@theckman could you help us on this?

snaveen:
@ranjib Could you help us on this?

PR #166, opened 2019-08-27
Correct JSON payload format for CreateIncident call

joepurdy:
Fixes https://github.com/PagerDuty/go-pagerduty/issues/165

Unlike other areas of the library the `CreateIncident` function failed to nest the options for the call in a top level `incident` JSON property.

This caused all calls to this function to fail API validation regardless of if they are structured correctly.


joepurdy:
@stmcallister Hey Scott! Hate to be a nag, but could you or another maintainer of go-pagerduty take a look at this?

The bug is causing some pain for folks since creating an incident (pretty important functionality!) is broken.

Happy to adjust my fix if there's feedback on a more preferred solution!

abtris:
Thanks for that PR!

PR #168, opened 2019-09-10
Use Go modules

nbutton23:
Simply ran `go mod init github.com/pagerduty/go-pagerduty`

Converted `PagerDuty/go-pagerduty` to match lowercased package nameing standards
https://blog.golang.org/package-names

Updated the `github.com/Sirupsen/logrus` import to current lowercase name


nbutton23:
@stmcallister Did you get a chance to look at this?

nbutton23:
Modules are used by basically every opensource go project. while its still beta (last i read they are targeting release in 1.14) its the most supported package manager out there for go.  Little has changed in the actual API of modules in the last few releases, so i would say is stable.

stmcallister:
I'm leaning toward @nbutton23 on this one. I'll take a look at this tomorrow.

stmcallister:
@nbutton23 -- We're still discussing this a bit internally. If I understand it correctly, going to modules will require all existing projects using this library to be moved out of the $GOPATH, which means everyone's projects will break once we release. Is there a smoother way to make this transition?

nbutton23:
@stmcallister It shouldn't break anything.  Right now by default GO111MODULE is set to auto.  If they are not using mods then nothing will change. 

```
enables module-mode if any go.mod is found, even inside GOPATH. (Prior to Go 1.13, GO111MODULE=auto would never enable module-mode inside GOPATH).
```
https://github.com/golang/go/wiki/Modules

stmcallister:
@nbutton23 Except that it broke for me. 😄 

```
$ make build
go get ./...
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
    ignoring go.mod;
    see 'go help modules'
command/addon_install.go:6:2: case-insensitive import collision: "github.com/pagerduty/go-pagerduty" and "github.com/PagerDuty/go-pagerduty"
make: *** [build] Error 1
```

I'm guessing this is because the package name changed from `PagerDuty` to `pagerduty`?

nbutton23:
Ah yeah thats not the module is self. Its the change made to match naming standards.  I can pull that out. . .

nbutton23:
@stmcallister Done!

stmcallister:
@nbutton23 Awesome! Thanks! LGTM.

Issue #160, opened 2019-07-23
Update logrus imports to github.com/sirupsen/logrus


ebracho:
The logrus owner recently changed the case of their handle (Sirupsen -> sirupsen). This causes issues on case-insensitive filesystems (like in macos), since vendoring both versions of the package results in filepath collisions. The logrus maintainer has stated that they will keep the new name (see https://github.com/sirupsen/logrus/issues/570), so it is up to other packages to update their imports.

fxfactorial:
I am getting this issue as well: 

```
 go build -o $GOPATH/bin/pd command/*
go: finding github.com/google/go-querystring/query latest
go: github.com/Sirupsen/logrus@v1.4.2: parsing go.mod: unexpected module path "github.com/sirupsen/logrus"
go: error loading module requirements
```

@stmcallister Can you please suggest a fix? I just want to just use `pd service list`

I looked at releases as well hoping that project would provide precompiled binaries (I need OS X) but it wasn't there either, i think that would be a good idea.

stmcallister:
Thanks for pointing this out! I'll take a look at this. But, it may not be until after the PagerDuty Summit conference next week.

stmcallister:
This was addressed with [PR 168](https://github.com/PagerDuty/go-pagerduty/pull/168).

PR #172, opened 2019-09-30
Change makefiles and readme

dineshba:
Tasks:
1) Introduce go mod
2) Update Makefile targets

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,6 @@ jobs:
   build:
     docker:
       - image: circleci/golang:1.12
-    working_directory: /go/src/github.com/PagerDuty/go-pagerduty
     steps:
       - checkout
       - run:
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 bin/*
 *.swp
+pd
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bin/*
 *.swp
 pd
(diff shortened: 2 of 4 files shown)


dineshba:
Fixes this issue also https://github.com/PagerDuty/go-pagerduty/issues/160

dineshba:
Hi @stmcallister Made the changes to the PR. I am okay with `GO111MODULE` being auto.

stmcallister:
Thanks, @dineshba! Also, any strong objections to adding the following lines to `.gitignore`?

```
vendor/
pd
```

dineshba:
@stmcallister Added `pd` to gitignore. 

Can we check in `vendor` folder to `git` ? Advantage: Builds will be very fast in CI(as vendor is already present with code). Whats your opinion on this ?

stmcallister:
@dineshba  We discussed this internally, and checking the `vendor` directory into the repo has pluses  and negatives. The build speed you mentioned would be nice, but the policing of dependencies would be  more tedious. Especially when developers can get the same benefit by running `go mod vendor`. So, we would rather not check the `vendor` directory into the repo, but instead either have devs run `go mod vendor` or include that in the makefile.

dineshba:
@stmcallister I agree with you. Added `vendor/` to `.gitignore`. As you suggested, developers can still use `go mod vendor` and build using `vendor` directory. So i didn't modify `Makefile`

PR #173, opened 2019-10-02
client: allow overriding the api endpoint

heimweh:
With this change in place, we'll be able to override the API endpoint when running tests.
Happy to follow up with a basic test example in this or a separate PR.


stmcallister:
LGTM. Thanks, @heimweh!

PR #174, opened 2019-10-03
Fix import of logger logrus to correct module path

KeisukeYamashita:
## What

* Fix `github.com/Sirupsen/logrus` to `github.com/sirupsen/logrus`

## Why

* The module path is not correct and the `command/` does not work.


stmcallister:
Thanks for submitting this! 

However, this issue was addressed with [PR 168](https://github.com/PagerDuty/go-pagerduty/pull/168) which was just merged.

PR #175, opened 2019-10-04
Adds support for calling the manage event API with a golang context

danmrichards:
# Summary
Adds a new `ManageEventWithContext` function which performs the same task as `ManageEvent` but allows the caller to supply a context. This context can then be used for request cancellation in the event of a deadline or timeout. 

`ManageEvent` now becomes a wrapper of `ManageEventWithContext` just passing a background context, the underlying functionality will not have changed.

# Reasoning
The current `ManageEvent` method does not allow this and could become problematic should the HTTP request run slowly. It is preferable to put this control in the hands of the caller, should they want to take advantage of this. If not they can continue to use `ManageEvent` as normal.

# Other Changes
* Fixed some missing error checking when building the HTTP request
* Use the HTTP method constant instead of a string literal
* Fixed package name collision with the `bytes` variable

### DIFF
--- a/event_v2.go
+++ b/event_v2.go
@@ -2,6 +2,7 @@ package pagerduty
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


theckman:
I want to get `context.Context` support into this package for all methods, and so I'm going to close this now and assume we'll align on a path forward in #267.

danmrichards:
Sure, thanks for the update @theckman

PR #182, opened 2019-10-17
Adding proper types for Links and Images

daniellockard:
I think this is correct? It should make code easier to write instead of writing something like 

```
Links: []interface{}{
			map[string]string{
				"href": "https://link.to.thing/thing",
				"text": "Link",
			},
		},
```

### DIFF
--- a/event_v2.go
+++ b/event_v2.go
@@ -10,14 +10,27 @@ import (
 
 // Event includes the incident/alert details
 type V2Event struct {
-	RoutingKey string        `json:"routing_key"`
-	Action     string        `json:"event_action"`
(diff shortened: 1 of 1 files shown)


theckman:
@daniellockard to confirm: would this is would result in a compile-time breaking API change?

daniellockard:
This PR was opened quite some time ago. I haven't actually looked at pagerduty anything in quite some time. I believe it would as it changes the types of images and links to actual types.

theckman:
@daniellockard thank you for taking the time to comment. That makes sense, and I'll keep this PR on-deck for whenever we look to release v2.0.0. Thank you for the contribution, and I apologize for it taking so long to get some sort of resolution on this.

PR #190, opened 2020-01-09
Tests

stmcallister:
This PR adds tests for each of the endpoints, with the exception of events_v2.


theckman:
Can you please break this in to smaller PRs to reviewing it is feasible?

go.mod
```diff
@@ -7,5 +7,6 @@ require (
 	github.com/mitchellh/cli v1.0.0
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/sirupsen/logrus v1.4.2
+	github.com/stretchr/testify v1.2.2
```
theckman: I would like to see us stick to the standard library as much as possible.

stmcallister: @theckman , something more like this? Where I use `reflect` for the comparison in an if statement. 

```
func TestAbility_ListAbilities(t *testing.T) {
	t.Parallel()
	setup()
	defer teardown()

	mux.HandleFunc("/abilities", func(w http.ResponseWriter, r *http.Request) {
		testMethod(t, r, "GET")
		w.Write([]byte(`{"abilities": ["sso"]}`))
	})

	var client = &Client{apiEndpoint: server.URL, authToken: "foo", HTTPClient: defaultHTTPClient}
	want := &ListAbilityResponse{Abilities: []string{"sso"}}

	res, err := client.ListAbilities()

	if err != nil {
		t.Fatal(err)
	}

	if !reflect.DeepEqual(res, want) {
		t.Errorf("returned %#v; want %#v", res, want)
	}
}
```

Also, as for the size of the PR, adding testing is no small task. But didn't seem I should do in pieces. In hindsight, might have been smarter to propose a testing style and then go forward.

stmcallister:
Hey @theckman. I went ahead and removed the `testify` dependency from the project.  Also, as mentioned before, I wanted to introduce testing by covering as many of the functions as possible, rather than introduce testing incrementally. The project needs testing, and I think this will help review and merge other PRs faster and with more confidence. Thoughts?

stmcallister:
Hey @theckman, do you have an idea of when you will you be able to review this? If you don't have the bandwidth, maybe we could tag another reviewer?

theckman:
@stmcallister sorry for the delay; was gone all of last week for yearly planning. Totally nuked me for the weekend. I'll do a review of this, and direct comments to you to be addressed later.

PR #192, opened 2020-01-27
List team members, single page or all (with helper for auto-pagination)

mwhite-ibm:
## What?

This PR adds:
* `ListMembers`: Method to get a single page from the `members` endpoint (`/teams/<ID>/members`).
* `pagedGet`: Helper for stepping through all available pages of a base endpoint `GET` request.
* `ListAllMembers`: Uses `pagedGet` to get _all_ members.

## Why?

Re. `ListMembers`: The `/teams/<ID>/members` endpoint is not currently accessible via the client package.
* This is helpful for retrieving team role information.

Re. `pagedGet` + `ListAllMembers`: All `List*` methods only return a single page of data.
* `pagedGet` allows one to make a single method call to get all members.
* This can be extended to the other `List*` methods to create `ListAll*`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -179,3 +179,38 @@ func (c *Client) getErrorFromResponse(resp *http.Response) (*errorObject, error)
 	}
 	return &s, nil
 }
+
+// responseHandler is capable of parsing a response. At a minimum it must
+// extract the page information for the current page. It can also execute
+// additional necessary handling; for example, if a closure, it has access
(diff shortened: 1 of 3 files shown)


mwhite-ibm:
@stmcallister : I've merged with `master` and modified my tests to use `setup` and `teardown`. 

Please let me know what other changes you would like, thanks!

mwhite-ibm:
@stmcallister Great, happy to contribute it!

PR #196, opened 2020-02-06
Implement the Incident endpoint for ResponderRequest

CerealBoy:
Was hoping to use this functionality, but it lacks in the API here, so adding it in. Based on [the docs](https://api-reference.pagerduty.com/#!/Incidents/post_incidents_id_responder_requests) this looks to be correct?

I wasn't entirely sure about the test case either, whether there is a way to provide better response content from the API (or have it generated).

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -380,4 +380,46 @@ type ListAlertResponse struct {
 	Alerts []Alert `json:"alerts,omitempty"`
 }
 
-/* TODO: Manage Alerts, Get Alert, Create Status Updates, Create Responder Request, */
+// CreateIncidentResponderRequestResponse
+type CreateIncidentResponderRequestResponse struct {
+	ResponderRequest ResponderRequest `json:"responder_request"`
+}
+
+// CreateIncidentResponderRequestOptions defines the input options for the Create Responder function.
+type CreateIncidentResponderRequestOptions struct {
+	From        string
+	Message     string
+	RequesterID string
+	Targets     []APIObject
+}
+
+// ResponderRequest contains the API structure for an incident responder request.
+type ResponderRequest struct {
+	Incident    Incident    `json:"incident"`
+	Requester   User        `json:"user,omitempty"`
+	RequestedAt string      `json:"request_at,omitempty"`
+	Message     string      `json:"message,omitempty"`
+	Targets     []APIObject `json:"responder_request_targets,omitempty"`
+}
+
+// CreateIncidentResponderRequest will submit a request to have a responder join an incident.
+func (c *Client) CreateIncidentResponderRequest(id string, o CreateIncidentResponderRequestOptions) (*CreateIncidentResponderRequestResponse, error) {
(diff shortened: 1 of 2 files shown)


stmcallister:
Cool! Thank you for submitting this pull request! I'll leave my comments inline!

incident.go
```diff
@@ -380,4 +380,46 @@ type ListAlertResponse struct {
 	Alerts []Alert `json:"alerts,omitempty"`
 }
 
-/* TODO: Manage Alerts, Get Alert, Create Status Updates, Create Responder Request, */
+// CreateIncidentResponderRequestResponse
+type CreateIncidentResponderRequestResponse struct {
```
stmcallister: I think we'll go with a shorter name. Will you remove all the `CreateIncident`s and start the names at `ResponderRequest`? So, this struct would be `ResponderRequestResponse`.

incident.go
```diff
@@ -380,4 +380,46 @@ type ListAlertResponse struct {
...
+
+// CreateIncidentResponderRequest will submit a request to have a responder join an incident.
+func (c *Client) CreateIncidentResponderRequest(id string, o CreateIncidentResponderRequestOptions) (*CreateIncidentResponderRequestResponse, error) {
+	data := make(map[string]interface{})
+	headers := make(map[string]string)
+	headers["From"] = o.From
+
+	data["message"] = o.Message
+	data["request_id"] = o.RequesterID
+	data["responder_request_targets"] = o.Targets
```
stmcallister: Rather than piece out each field in the payload object, will you create a `ResponderRequestPayload` struct with json tags and add that to the `ResponderRequestOptions`?

incident_test.go
```diff
@@ -341,3 +341,38 @@ func TestIncident_ListLogEntries(t *testing.T) {
...
 }
+
+func TestIncident_ResponderRequest(t *testing.T) {
+	setup()
+	defer teardown()
+
+	id := "1"
+	mux.HandleFunc("/incidents/"+id+"/responder_requests", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+		w.Write([]byte(`{"requester_id": "PL1JMK5", "message": "Help", "responder_request_targets": [{"responder_request_target":{"id":"PJ25ZYX","type":"user_reference"}}]}`))
```
stmcallister: This is supposed to match the JSON of the response you're expecting. Can you change it to reflect the response?

CerealBoy:
Hi @stmcallister, thank you very much for the feedback, I've made the changes as you've requested. The only difference I have is to use the ``ResponderRequestOptions`` as the payload, rather than having another struct purely for it. Would you prefer this to be split out?

Looking at the original commit, I can't work out how I managed to think the response payload looked the way it did considering the docs are completely different, so I've made a bunch of changes to better model the expected response data from this request.

stmcallister:
Looks good! Thanks for the feature and for making those changes @CerealBoy !

PR #197, opened 2020-02-12
Adding User-Agent Headers

stmcallister:
Adding User-Agent header to calls so that we understand how often the library is used. Also, rolling back #183. We don't want to be supporting undocumented features in this library. And, there are no immediate plans to document `oncall` for the escalation_policy endpoint.

### DIFF
--- a/client.go
+++ b/client.go
@@ -140,6 +140,7 @@ func (c *Client) do(method, path string, body io.Reader, headers *map[string]str
 			req.Header.Set(k, v)
 		}
 	}
+	req.Header.Set("User-Agent", "go-pagerduty")
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Authorization", "Token token="+c.authToken)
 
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -140,6 +140,7 @@ func (c *Client) do(method, path string, body io.Reader, headers *map[string]str
 			req.Header.Set(k, v)
 		}
 	}
+	req.Header.Set("User-Agent", "go-pagerduty")
```
dobs: It'd be good to add a version here as well, even just hardcoding `go-pagerduty/0.0.0` for the time being. Then it's at least something we can parse for.

Hardcoding because I think we need to fix how we track versioning -- the value in `main.go` doesn't match our tagged version and isn't super accessible from `client.go`.

event_v2.go
```diff
@@ -48,6 +48,7 @@ func ManageEvent(e V2Event) (*V2EventResponse, error) {
 		return nil, err
 	}
 	req, _ := http.NewRequest("POST", v2eventEndPoint, bytes.NewBuffer(data))
+	req.Header.Set("User-Agent", "go-pagerduty")
```
dobs: Ditto here on specifying a version.

PR #201, opened 2020-02-19
Handle Incident Log Entries of User Notification

joez-tkpd:
Add `User` struct on `LogEntries`

Because on Incident's Log Entries `/incidents/{id}/log_entries` when `"type": "notify_log_entry"`
the payload doesn't contain any information about the notified user except on user object

additionally add `Service` on `LogEntries`
additionally `Channel` also missed `Notification` object

### DIFF
--- a/log_entry.go
+++ b/log_entry.go
@@ -11,7 +11,8 @@ type Agent APIObject
 
 // Channel is the means by which the action was carried out.
 type Channel struct {
-	Type string
+	Type         string
+	Notification Notification
 }
 
 // Context are to be included with the trigger such as links to graphs or images.
@@ -29,11 +30,13 @@ type LogEntry struct {
 	CreatedAt              string `json:"created_at"`
 	Agent                  Agent
 	Channel                Channel
+	Service                Service
 	Incident               Incident
 	Teams                  []Team
 	Contexts               []Context
 	AcknowledgementTimeout int `json:"acknowledgement_timeout"`
 	EventDetails           map[string]string
+	User                   User
 }
 
 // ListLogEntryResponse is the response data when calling the ListLogEntry API endpoint.


stmcallister:
Thanks for submitting this PR! As a rule, we're not adding support for any undocumented API features. And, as of right now, none of these fields are documented in the PagerDuty API Reference. I'm working with the engineering team that owns this endpoint to see if we can change that. Will keep you posted. 

Until then, the `Service` and `User` fields should probably be of type `APIObject` as that type more accurately represents the objects on the `log_entries` endpoint.

obl1v1us:
Hi @stmcallister ! I'm also interested in retrieving User data from the LogEntries. Has there been any progress getting this added to the library? Or are we still waiting on having the Pagerduty API reference updated before adding this change?

theckman:
@joez-tkpd @obl1v1us I'm not a PD employee so I don't have any inside baseball, but it seems like the API documentation still has not been updated to include these fields:

https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1incidents~1%7Bid%7D~1log_entries/get

This branch is also conflicting with `master` and cannot be merged in its current form so we would need to get it updated, and it may also want to be merged with #264 as it seems like the `Channel` type (which is what's conflicting) is not correct in `master`.

@stmcallister can you speak to the documentation status?

theckman:
@joez-tkpd with #264 merged, some of what you're trying to accomplish is now possible. The `Channel` type has a `Raw` field on it that's a `map[string]interface{}` of all the fields (except for `type`) sent via JSON. You should be able to start using this once we release `v1.4.0`.

I'll leave the question about the API documentation for Scott.

theckman:
Ah I found it.

stmcallister:
Hey everyone! Sorry for taking so long on this one. After trying to get one of the engineering teams to update the PagerDuty API Docs to match this PR I took matters into my own hands and figured out how to do it myself. The PR has been approved and should be merged in the coming days. Thanks for your patience!

stmcallister:
[Log Entry Docs](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE1NA-list-log-entries) are updated.

theckman:
@joez-tkpd this will be part of v1.5.0 when released (merged via #377).

PR #208, opened 2020-03-03
Add Teams to Schedule

miekg:
Nor the API docs, nor the current code says so, but Schedule can take a
team (or more than one). Add this by using an APIReference slice.

Signed-off-by: Miek Gieben <miek@miek.nl>


stmcallister:
Thanks for the PR @miekg ! As a rule, we're only including documented features in our code libraries. That said, your PR gave me the avenue to poke our engineering team to add this to the docs! :) When the docs are updated I will merge. Thanks!

PR #214, opened 2020-04-09
List incident alerts

kilianw:
Add query options to the 'List alerts for an incident' - /incidents/{id}/alerts


stmcallister:
@kilianw - Thanks for making this contribution! Adding `ListIncidentAlertsOptions` will certainly add values to users. However, adding the additional argument to the `ListIncidentAlerts` function will break that call for existing users of the function. 

How about we create a `ListIncidentAlertsWithOpts` function so that we remain backwards compatible? We could even have `ListIncidentAlerts` call the new options function to avoid duplicate code.

PR #225, opened 2020-06-02
Update incident: fix FirstTriggerLogEntry type

afarbos:
Use LogEntry type instead of APIObject for FirstTriggerLogEntry attribute in Incident


afarbos:
Pagerduty does return a LogEntry, not just an APIObject when listing incident. This PR will expose the full object and not just the APIObject.

PR #229, opened 2020-06-17
Add pagination options to get rulesets

kilianw:
Add options to enable pagination for list rulesets operation.

### DIFF
--- a/ruleset.go
+++ b/ruleset.go
@@ -2,6 +2,7 @@ package pagerduty
 
 import (
 	"fmt"
+	"github.com/google/go-querystring/query"
 	"net/http"
 )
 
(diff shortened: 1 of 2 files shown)


stmcallister:
Thanks for submitting this, and sorry for the delay in response. I like the  idea of adding the option to send in options while getting Rulesets. However, why was the ability to list all the  rulesets removed? I see the function is still there, but it's not going through all the pages and gather all the rulesets into a single list.

ruleset.go
```diff
@@ -132,37 +133,24 @@ type RuleActionExtraction struct {
...
+// ListRulesetsWithOpts get all rulesets with options
+func (c *Client) ListRulesetsWithOpts(o APIListObject) (*ListRulesetsResponse, error) {
+	v, err := query.Values(o)
+	if err != nil {
+		return nil, err
 	}
-
-	// Make call to get all pages associated with the base endpoint.
-	if err := c.pagedGet("/rulesets/", responseHandler); err != nil {
+	resp, err := c.get("/rulesets?" + v.Encode())
```
theckman, line 142: Because this removes the pagination it would be a breaking API change, can you revert back to using `pagedGet()`?

theckman (changes requested):
One requested change, and it looks like this will need a merge conflict resolved before we can merge it.

ruleset.go
```diff
@@ -132,37 +133,24 @@ type RuleActionExtraction struct {
...
-
-		// Return stats on the current page. Caller can use this information to
-		// adjust for requesting additional pages.
-		return APIListObject{
-			More:   result.More,
-			Offset: result.Offset,
-			Limit:  result.Limit,
-		}, nil
+// ListRulesetsWithOpts get all rulesets with options
+func (c *Client) ListRulesetsWithOpts(o APIListObject) (*ListRulesetsResponse, error) {
```
theckman, line 137: I also forgot, we probably want to update this method to accept a `context.Context` as its first argument, since both `.get()` and `.pagedGet()` now require `context.Context` as their first argument.

theckman: @kilianw we've aligned on moving forward with adding `context.Context` support to new methods going forward. So before merging we'd need to update this method to take `ctx context.Context` as its first argument, to then pass them to `pagedGet()`. Thanks!

theckman:
So I'm now realizing that we should probably just close this PR. The method being changed doesn't need to take these options because it's handling the pagination for you. The return type chosen for this method isn't appropriate, because it makes it seem like it's the response for one of the requests and not all of them. We cannot change that as it's a breaking change, but it's on the books for v2.

In the mean time, I've made it more clear that this gives you all of the Rulesets by changing the signature of the method when we added `context.Context` support. In v2, the method will be changed to accomplish what you were trying to.

Thank you again for your contribution, I'll close this PR a this time.

PR #234, opened 2020-07-29
allow creating services with no scheduled actions

cluarkhpe:
When using `use_support_hours` as the type for `incident_urgency_rule` you have to specify the `scheduled_actions` array as well.
The PagerDuty API allows this array to be empty, but an empty array will be omitted from the JSON.
This causes API errors when trying to create services that notify based on support hours but don't have any scheduled actions.


stmcallister (approved):
LGTM! Thanks for submitting this! 🎉

PR #235, opened 2020-08-17
Adding metadata to services, which are present in V2 webhooks

tlianza:
Metadata is present on Incidents *and* Services when you receive a webhook. Without this, you don't get the conference_url, etc for your service.

### DIFF
--- a/service.go
+++ b/service.go
@@ -74,6 +74,7 @@ type Service struct {
 	AlertCreation          string               `json:"alert_creation,omitempty"`
 	AlertGrouping          string               `json:"alert_grouping,omitempty"`
 	AlertGroupingTimeout   *uint                `json:"alert_grouping_timeout,omitempty"`
+	Metadata               interface{}          `json:"metadata,omitempty"`
 }
 
 // ListServiceOptions is the data structure used when calling the ListServices API endpoint.


tlianza:
@stmcallister - what do you think?

service.go
```diff
@@ -74,6 +74,7 @@ type Service struct {
 	AlertCreation          string               `json:"alert_creation,omitempty"`
 	AlertGrouping          string               `json:"alert_grouping,omitempty"`
 	AlertGroupingTimeout   *uint                `json:"alert_grouping_timeout,omitempty"`
+	Metadata               interface{}          `json:"metadata,omitempty"`
```
theckman, line 77: Are we able to be more precise than `interface{]`?

tlianza: It doesn't looks like it's any more specific in the IncidentDetails itself: https://github.com/PagerDuty/go-pagerduty/blob/master/webhook.go#L30 . I assumed in all cases they're using this object pretty freeform.

stmcallister: @tlianza Thanks for adding this! As a rule we're only adding API features that are documented in the PagerDuty API Docs and the Metadata fields isn't listed. I'll contact that engineering team internally and see if there are plans to document it.

theckman: @stmcallister it is shown in the example JSON document on this page:

https://developer.pagerduty.com/docs/webhooks/v2-overview/

What's interesting is that there is a `metadata` field documented, but it's shown as part of the `incident` object and not the `service` object and its comment seems service related: "The metadata saved on the service."

It's extremely bizarre that you'll show that the fields exist in the example JSON, but not document them and refuse to support them. Why even show them in the example JSON then?

theckman:
It doesn't seem like PagerDuty has updated the documentation for V2 Webhooks, and with them being deprecated I suspect they might not. It may be best to close this issue and see whether we can support this for V3 webhooks (assuming they are documented).

stmcallister:
My apologies for letting this one slip. @theckman is correct. With the introduction of [Webhooks V3](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTkw-v3-overview) the engineering teams won't be updating Webhooks V2. 

In looking at the V3 docs, the `conference_url` is documented in the [Event Data Types](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTkw-v3-overview#event-data-types) section of the Webhook docs.

theckman:
Considering there is issue #367 to track adding V3 Webhook support into this library, and V2 Webhooks are deprecated and won't be getting any updates, I'm going to mark this PR as closed without merging the change.

Thank you so much for your contribution, and I'm sorry it took so long to come to a final resolution.

PR #236, opened 2020-08-27
Fix ruleset rule not respecting position "0"

zane-deg:
During the JSON marshal of a `RulesetRule`, a `Position` with the value of `0` is considered empty and is not sent in the POST or PUT payload. Using a pointer allows the JSON marshal function to distinguish between an empty value and a default value.

I'm not sure how this change would be handled as it isn't really backwards compatible. Happy to hear any thoughts or opinions.

### DIFF
--- a/ruleset.go
+++ b/ruleset.go
@@ -39,7 +39,7 @@ type ListRulesetsResponse struct {
 // RulesetRule represents a Ruleset rule
 type RulesetRule struct {
 	ID         string          `json:"id,omitempty"`
-	Position   int             `json:"position,omitempty"`
+	Position   *int            `json:"position,omitempty"`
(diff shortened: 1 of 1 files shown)


theckman:
Yeah that's a bummer, but understandable. If it weren't such a big deal to have consumers upgrade to v2.0.0, I'd be OK with merging this and starting the v2 branch a la Chrome.

What we may need to do is cut a breaking minor release, and treat it like a necessary 1.0 bugfix.

dobs:
@theckman Agreed that a minor release sounds reasonable.

I believe there are other breaking changes we'd want to incorporate into a full V2 release, and at the very least a more thorough pass on properly handling optional fields.

Barring any objections I'll cut v1.3.0 later today.

dobs (approved):

PR #237, opened 2020-08-31
Add assignees to log entry

tautvydass:
'Assign' log entries have assignees field.

### DIFF
--- a/log_entry.go
+++ b/log_entry.go
@@ -35,6 +35,7 @@ type CommonLogEntryField struct {
 	Contexts               []Context         `json:"contexts,omitempty"`
 	AcknowledgementTimeout int               `json:"acknowledgement_timeout"`
 	EventDetails           map[string]string `json:"event_details,omitempty"`
+	Assignees              []APIObject       `json:"assignees,omitempty"`
 }
 
 // LogEntry is a list of all of the events that happened to an incident.


log_entry.go
```diff
@@ -35,6 +35,7 @@ type CommonLogEntryField struct {
 	Contexts               []Context         `json:"contexts,omitempty"`
 	AcknowledgementTimeout int               `json:"acknowledgement_timeout"`
 	EventDetails           map[string]string `json:"event_details,omitempty"`
+	Assignees              []APIObject       `json:"assignees,omitempty"`
```
theckman, line 38: @tautvydass I've seen @stmcallister make earlier comments about not wanting to accept changes to this library that rely on undocumented API features. Looking through the API docs I don't see this field documented, and so I worry that this might not be accepted.

@stmcallister would you be able to confirm that?

stmcallister: In hunting around, I've discovered this is [obscurely documented](https://developer.pagerduty.com/api-reference/c2NoOjI3NDc5OTk-assign-log-entry). I'll take it.

stmcallister (approved):
Thanks for doing this and for your patience! 🎉 🎉 🎉

theckman:
@stmcallister happy to merge this, but I don't yet see this in the docs on the site:

https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE1NA-list-log-entries
https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE1NQ-get-a-log-entry

theckman:
Found it: https://developer.pagerduty.com/api-reference/c2NoOjI3NDc5OTk-assign-log-entry

PR #241, opened 2020-09-11
Event v2 client

goatherder:
Implements ManageEvent for the pagerduty Client - so we can mock it out and test it.


stmcallister (approved):
This looks great! Thank you for contributing these changes!

PR #233, opened 2020-07-16
Couple of syntax and unmarshaling fixes. closes #232

Erog38:
This MR adds a field to the webhook struct, fixes struct tag syntax, and adjusts the payload for sending responder requests to match the documentation. Closes #232

PR #242, opened 2020-09-12
Fix struct tags

bvwells:
Fix struct tags. Also clean up some documentation.


bvwells:
Just noticed that there is already a PR for this #233 so closing.

PR #246, opened 2020-10-09
Add Support For Change Events

Sjeanpierre:
This PR implements the newly released [Change Events](https://developer.pagerduty.com/docs/events-api-v2/send-change-events/#try-it-out) feature. PR #241 seems to have taken a better approach on dealing with V2 Event endpoints in general, so if we can get some consensus on that approach I'd be more than happy to refactor that bit of my implementation.

### DIFF
--- a/change_events.go
+++ b/change_events.go
@@ -19,10 +19,10 @@ type ChangeEvent struct {
 // Payload ChangeEvent Payload
 // https://developer.pagerduty.com/docs/events-api-v2/send-change-events/#example-request-payload
 type Payload struct {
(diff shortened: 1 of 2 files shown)


joelcrocker:
:+1: for the approach for dealing with events API endpoints from #241.

Sjeanpierre:
Refactored to match the style from PR #241 should be good for review now @stmcallister

joelcrocker (approved):
:+1:

stmcallister (approved):
Great commenting and thanks for adding tests!

PR #248, opened 2020-10-27
Correct namespacing of Change Event resource

Sjeanpierre:
In reviewing my previous implementation I noticed that some structs had more global namespaces that strictly necessary which can be non-intuitive when looking at the entire package in aggregate.  For example the exported Payload struct didn't make much sense in its stand alone form, so this have been updated to ChangeEventPayload as to avoid confusion and namespace collisions. Same applies to ChangeEvent links.

### DIFF
--- a/change_events.go
+++ b/change_events.go
@@ -6,28 +6,28 @@ import (
 	"errors"
 )
 
-const ChangeEventPath = "/v2/change/enqueue"
+const changeEventPath = "/v2/change/enqueue"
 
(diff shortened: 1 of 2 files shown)


joelcrocker (approved):
:+1:

stmcallister (approved):
Normally, I wouldn’t be a huge fan of changing object names. However, since this one has only been out a short time, I don’t think it will affect many users.

Thanks for the improvement!!!

PR #253, opened 2020-12-02
business_service: clean b.ID before calling UPDATE

GiedriusS:
PUT `/business_services/{id}` doesn't accept `ID` in the body:
https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1business_services~1%7Bid%7D/put

However, specifying it for the `go-pagerduty` to work is mandatory.
Thus, it is impossible to use this method ATM without this fix. Let's
clear `b.ID` before calling `PUT` and store it in a separate variable.

Testing: my code calling this method works again with these changes.
Otherwise, the following error is spotted:
```
Failed call API endpoint. HTTP response code: 400. Error: \u0026{400 Invalid input [The field id is invalid.]}
```

I have added a small test case to the suite. LMK if this looks good.

### DIFF
--- a/business_service.go
+++ b/business_service.go
@@ -106,8 +106,10 @@ func (c *Client) DeleteBusinessService(ID string) error {
 // UpdateBusinessService updates a business_service.
 func (c *Client) UpdateBusinessService(b *BusinessService) (*BusinessService, *http.Response, error) {
 	v := make(map[string]*BusinessService)
+	id := b.ID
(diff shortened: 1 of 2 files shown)


GiedriusS:
@stmcallister could you please look into this?

stmcallister (approved):
Good catch! Thanks for fixing that!! 🎉

PR #254, opened 2020-12-08
service_dependency: fix (dis-)associate operations

GiedriusS:
Fix the (dis-)associate functions by not doing the hocus pocus with
`map[string]...`. `relationships` already exists due to the JSON tag in
the struct. Without this, the payload looks like:

```json
{"relationships": {"relationships": [...] } }
```

And PD API returns "invalid request". Tested locally that it works as
expected with these changes.


stmcallister (approved):
Another great catch. Thank you!! 🎉

PR #255, opened 2020-12-09
fix creating services that don't use support hours

cluarkhpe:
It appears that when creating a service with `scheduled_actions`: `null`, the API requires that `support_hours` is also set to `null`.
Without this change the API returns an error: `HTTP response code: 400. Error: &{2001 Invalid Input Provided [Support hours is required.]}`

### DIFF
--- a/service.go
+++ b/service.go
@@ -69,7 +69,7 @@ type Service struct {
 	EscalationPolicy        EscalationPolicy         `json:"escalation_policy,omitempty"`
 	Teams                   []Team                   `json:"teams,omitempty"`
 	IncidentUrgencyRule     *IncidentUrgencyRule     `json:"incident_urgency_rule,omitempty"`
-	SupportHours            *SupportHours            `json:"support_hours,omitempty"`
+	SupportHours            *SupportHours            `json:"support_hours"`
 	ScheduledActions        []ScheduledAction        `json:"scheduled_actions"`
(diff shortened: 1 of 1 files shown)


stmcallister:
Hello! Can you share more context of the scenario you're describing? I'm unable to reproduce the behavior on the API. When I remove `scheduled_actions` by setting it to `null` as you described I'm still able to create a service with `support_hours` set.

GiedriusS:
Heh, created a PR for the same issue: https://github.com/PagerDuty/go-pagerduty/pull/258. @stmcallister I think you got this wrong - we are talking about `support_hours` here. We don't want to set any value for it or IOW we want to use whatever the defaults are. Having `omitempty` means that if the PTR is `nil` then it does _not_ end up in the JSON that we send over the wire. And the API rejects requests where this key does not exist - it must be either set to `null` explicitly or set to the actual value. PTAL again.

stmcallister:
@GiedriusS Oh! I see what you're saying. I'll take another look.

stmcallister:
@cluarkhpe There's been a change made to `service.go` since you created this PR, causing a conflict. Will you resolve the conflict. Once it's resolved I'll merge your change.

cluarkhpe:
@stmcallister should be good now

stmcallister (approved):
Awesome! Thank you for doing this 🎉 👍

PR #256, opened 2020-12-10
ruleset: remove omitempty from Route

GiedriusS:
Remove `omitempty` from Route because otherwise it is impossible to
remove a route from the catch-all rule by calling UpdateRulesetRule().


stmcallister (approved):
LGTM! 🎉 👍

PR #258, opened 2020-12-18
service: do not omit support hours

GiedriusS:
CreateService() API contract has been randomly changed without a notice
to not accept the non-existence of the `support_hours` key. Creating a
service currently fails with:

```
Failed call API endpoint. HTTP response code: 400. Error: \u0026{2001 Invalid Input Provided [Support hours is required.]}
```

The API gladly accepts the same payload when `support_hours` is
_explicitly_ set to `null` in the resulting JSON.


GiedriusS:
Actually, this has been done already: https://github.com/PagerDuty/go-pagerduty/pull/255. Let's merge that one!

PR #260, opened 2021-01-11
add ListServicesPaginated to services api (and tests)

c6h12o6:
Added new functionality to easily leverage pagedGet() in ListServices. Does not alter any existing functionality.

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+  "regexp"
 	"io"
 	"net"
 	"net/http"
@@ -247,6 +248,19 @@ func (c *Client) getErrorFromResponse(resp *http.Response) (*errorObject, error)
 	return &s, nil
 }
 
+// Helper function to determine wither additional parameters should use ? or & to append args
+func getBasePrefix(basePath string) string {
+ var conjunction string
+  // check to see if there are already arguments in basePath (look for ? in path)
+  re := regexp.MustCompile(".*/[^/]*\\?[^/]*$")
+  if re.Match([]byte(basePath)) {
+    conjunction = "&"
+  } else {
+    conjunction = "?"
+  }
+  return fmt.Sprintf("%s%s", basePath, conjunction)
+}
+
 // responseHandler is capable of parsing a response. At a minimum it must
 // extract the page information for the current page. It can also execute
 // additional necessary handling; for example, if a closure, it has access
@@ -261,9 +275,10 @@ func (c *Client) pagedGet(basePath string, handler responseHandler) error {
 	// Offset to set for the next page request.
 	var nextOffset uint
 
+  basePrefix := getBasePrefix(basePath)
 	// While there are more pages, keep adjusting the offset to get all results.
 	for stillMore, nextOffset = true, 0; stillMore; {
-		response, err := c.do("GET", fmt.Sprintf("%s?offset=%d", basePath, nextOffset), nil, nil)
+		response, err := c.do("GET", fmt.Sprintf("%soffset=%d", basePrefix, nextOffset), nil, nil)
 		if err != nil {
 			return err
 		}
--- a/client.go
+++ b/client.go
@@ -8,7 +8,7 @@ import (
 	"io"
 	"net"
 	"net/http"
-  "path"
+	"path"
 	"runtime"
 	"strings"
 	"time"
@@ -359,10 +359,10 @@ func (c *Client) getErrorFromResponse(resp *http.Response) APIError {
 
 // Helper function to determine wither additional parameters should use ? or & to append args
 func getBasePrefix(basePath string) string {
-  if strings.Contains(path.Base(basePath), "?") {
-    return basePath + "&"
-  }
-  return basePath + "?"
+	if strings.Contains(path.Base(basePath), "?") {
+		return basePath + "&"
+	}
+	return basePath + "?"
 }
 
 // responseHandler is capable of parsing a response. At a minimum it must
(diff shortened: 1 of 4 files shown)


service.go
```diff
@@ -120,6 +120,33 @@ func (c *Client) ListServices(o ListServiceOptions) (*ListServiceResponse, error
...
+
+		services = append(services, result.Services...)
+
+		return APIListObject{
+			More:   result.More,
+			Offset: result.Offset,
+			Limit:  result.Limit,
+		}, nil
+	}
+	if err := c.pagedGet("/services?"+v.Encode(), responseHandler); err != nil {
```
theckman, line 144: `pagedGet()` now takes a `context.Context` as its first argument:

https://github.com/PagerDuty/go-pagerduty/blob/1df5787177c5e24a3b886dbf22d67e903327aab7/client.go#L367

service.go
```diff
@@ -120,6 +120,33 @@ func (c *Client) ListServices(o ListServiceOptions) (*ListServiceResponse, error
 	return &result, c.decodeJSON(resp, &result)
 }
 
+// ListServices lists existing services processing paginated responses
+func (c *Client) ListServicesPaginated(o ListServiceOptions) ([]Service, error) {
```
theckman, line 124: Could you make the signature `ListServicesPaginated(ctx context.Context, o ListServiceOptions)` and pass the `ctx` into `pagedGet()`?

client.go
```diff
@@ -247,6 +248,19 @@ func (c *Client) getErrorFromResponse(resp *http.Response) (*errorObject, error)
 	return &s, nil
 }
 
+// Helper function to determine wither additional parameters should use ? or & to append args
+func getBasePrefix(basePath string) string {
+	var conjunction string
+	// check to see if there are already arguments in basePath (look for ? in path)
+	re := regexp.MustCompile(".*/[^/]*\\?[^/]*$")
```
theckman, line 255: This currently recompiles the entire regular expression for each method call that uses `pagedGet()`, at a minimum it should probably be a package-level unexported variable so that it can be reused.

client.go
```diff
@@ -247,6 +248,19 @@ func (c *Client) getErrorFromResponse(resp *http.Response) (*errorObject, error)
...
+	if re.Match([]byte(basePath)) {
+		conjunction = "&"
+	} else {
+		conjunction = "?"
+	}
+	return fmt.Sprintf("%s%s", basePath, conjunction)
+}
```
theckman, line 262: I think regular expressions make sense in systems where the filters / matches need to be user-supplied and cannot be known at build time. In this case, separate from my reuse comment, I'm not sure a regular expression is the right fit here. 

How would you feel about something like this? I've ran it against your proposed tests and got a :+1: :

```go
func getBasePrefix(basePath string) string {
	var p bool

	for _, c := range path.Base(basePath) {
		if c == '?' {
			p = true
			break
		}
	}

	if p {
		return basePath + "&"
	}

	return basePath + "?"
}
```

theckman (changes requested):
I think overall this is something we'd like to get merged in. There are some merge conflicts that need resolved, as well as some other non-conflicts that need updating, before we could merge this in.

Please let me know if you have any questions or comments on my feedback. Thank you for the contribution! 👍

client.go
c6h12o6: I take your point on the regular expression, but is there a reason you didnt just use:

if strings.Contains(path.Base(basePath), "?")) { ... }

?

c6h12o6:
Comments addressed @theckman, I used the strings.Contains() approach from above, but if theres a reason to iterate manually, I can do that too.

client.go
c6h12o6: Obsolete due to other changes.

service.go
c6h12o6: done.

client.go
theckman: ryan reynolds confused

Nope, wasn't thinking. 😄

service.go
```diff
@@ -121,6 +121,33 @@ func (c *Client) ListServices(o ListServiceOptions) (*ListServiceResponse, error
 	return &result, c.decodeJSON(resp, &result)
 }
 
+// ListServices lists existing services processing paginated responses
+func (c *Client) ListServicesPaginated(ctx context.Context, o ListServiceOptions) ([]Service, error) {
+	services := make([]Service, 0)
```
theckman, line 126: I totally missed this before because we don't have automatic linter checks turned on yet, but my editor flagged this. Mind converting this to `var services []Service`? The `append()` below will take care of allocation of the slice.

service_test.go
```diff
@@ -44,6 +47,59 @@ func TestService_List(t *testing.T) {
...
+	var client = &Client{apiEndpoint: server.URL, authToken: "foo", HTTPClient: defaultHTTPClient}
+	var opts = ListServiceOptions{
+		APIListObject: listObj,
+		TeamIDs:       []string{},
+		TimeZone:      "foo",
+		SortBy:        "bar",
+		Query:         "baz",
+		Includes:      []string{},
+	}
+	res, err := client.ListServicesPaginated(context.TODO(), opts)
```
theckman, line 83: Sorry I missed this too, mind changing this to a `context.Background()`?

theckman (changes requested):
I'm sorry, I noticed two more minor things and then this is ready for merging!

service_test.go
c6h12o6: You didn't miss this one - it's new code :-). I wasnt sure what context.Background does in a test, but sure.

service.go
c6h12o6: I think they're equivalent. Is there a style guide you're following?

c6h12o6:
@theckman done once again.

service.go
theckman: Not a specific style guide, just a set of linters I've been running on my stuff for years. Never had an issue with it so I tend to fix things it yells about. 

Thank you for the fix. :bow:

theckman (approved):

theckman:
@c6h12o6 git is awful and caught a merge conflict based on another PR we merged in. I'll fix it on merge in just a moment.

theckman:
@c6h12o6 thank you for the contribution!

PR #262, opened 2021-01-20
team: allow specifying role

GiedriusS:
Permit specifying the role when adding a user to a team. Tested locally. The role must be "manager", "observer", or "responder".

### DIFF
--- a/team.go
+++ b/team.go
@@ -26,6 +26,13 @@ type ListTeamOptions struct {
 	Query string `url:"query,omitempty"`
 }
 
+// AddUserToTeamOptions are the input parameters used when calling the AddUserToTeam API endpoint.
+type AddUserToTeamOptions struct {
+	TeamID string `json:"-"`
+	UserID string `json:"-"`
+	Role   string `json:"role,omitempty"`
+}
+
 // ListTeams lists teams of your PagerDuty account, optionally filtered by a search query.
 func (c *Client) ListTeams(o ListTeamOptions) (*ListTeamResponse, error) {
 	v, err := query.Values(o)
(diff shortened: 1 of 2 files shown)


team.go
```diff
@@ -26,6 +26,13 @@ type ListTeamOptions struct {
 	Query string `url:"query,omitempty"`
 }
 
+// AddUserToTeamOptions are the input parameters used when calling the AddUserToTeam API endpoint.
+type AddUserToTeamOptions struct {
+	TeamID string `json:"-"`
+	UserID string `json:"-"`
+	Role   string `json:"role,omitempty"`
```
theckman, line 33: Based on your comment in the PR, there are only a handful of values that are accepted. Instead of this being a `string`, should we make it a named `string` type so that we can have a set of constants that people can also use?

```go
type UserRole string

const (
  UserRoleManager UserRole = "manager"
  // ...
)
```

The `User` struct also has a `Role` field that's a `string`. I'm not sure if it makes sense for these to be different. 🤔 

To be clear, this comment isn't a request to change but looking for your thoughts / feedback.

team.go
```diff
@@ -84,8 +91,8 @@ func (c *Client) RemoveUserFromTeam(teamID, userID string) error {
 }
 
 // AddUserToTeam adds a user to a team.
-func (c *Client) AddUserToTeam(teamID, userID string) error {
-	_, err := c.put("/teams/"+teamID+"/users/"+userID, nil, nil)
+func (c *Client) AddUserToTeam(options AddUserToTeamOptions) error {
```
theckman, line 94: This modification makes this a breaking change, which will add some delays in merging.

How would you feel about creating a new method, and update `AddUserToTeam()` to call that one and indicate the other should be used instead? Then we can file an issue to clean this up in `v2.0.0`.

theckman (changes requested):

theckman:
I've taken inspiration from this PR, and added an option struct with `role` support to #308 on the new method that also accepts the `context.Context`. #308 will accomplish the same thing as this PR, without the breaking change, and so I'm going to close this PR in favor of that one.

Thank you for this contribution and the inspiration.

PR #263, opened 2021-01-22
Add capability to update incidents

bendehaan:
Add a method and corresponding unit test for updating incidents.

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -186,6 +186,25 @@ func (c *Client) CreateIncident(from string, o *CreateIncidentOptions) (*Inciden
 	return &ii.Incident, nil
 }
 
+// UpdateIncident updates an existing incident.
+func (c *Client) UpdateIncident(id string, i Incident) (*Incident, error) {
+	v := make(map[string]Incident)
+	v["incident"] = i
+	resp, e := c.put("/incidents/"+id, v, nil)
+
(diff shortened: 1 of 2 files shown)


bendehaan:
I understood from support that currently updating incidents is not supported. Is that something you will support in the future?

theckman:
@bendehaan could you share more about that context of updating not being supported? This documentation makes it seem like it is: https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1incidents~1%7Bid%7D/put

Full disclosure while I do have merge access I am not an employee, so the context you heard would be helpful if you could share it.

bendehaan:
@theckman Sure! I thought the same from documentation, but this is the message I got from PagerDuty tech support:
```
Hello Ben,

Thanks for reaching out. Once an Incident has been triggered, it can’t be updated. You would have to create a new alert and have this deduplicated or grouped into the incident to keep all the context in one place. But updating custom details ect within on Incident and Alert is not supported.

Kind regards, 
```

theckman:
@bendehaan thank you for that context. Yeah, I'd say it's probably best to close the PR since it's not something their API is supporting today. I don't have a great way to know if they ever will support that, so I'll close this PR for now. But please reopen it if that API support is ever added.

PR #264, opened 2021-01-28
Fix logentry.channel json marshaling/unmarshaling

StupidScience:
Hello.

I've faced an issue when unmarshalling and marshalling back of LogEntry lead to results that differ from initial.

First of all just because of absent json tags, so it lead to `"Type"` instead of `"type"`, so the next unmarshaling was broken.
Also "fixing" tags leads to nested object and
```json
{
    ...
    "channel": {
	    "type": "web_trigger",
	    "summary": "My new incident",
	    "details_omitted": false
    }
}
```

transforms to
```json
{
    ...
    "channel": {
	    "type": "web_trigger",
            "raw": {
	        "summary": "My new incident",
	        "details_omitted": false
            }
    }
}
```

So this PR will fix this, however it require Channel type changes.

Looking forward for your feedback.

### DIFF
--- a/log_entry.go
+++ b/log_entry.go
@@ -115,7 +115,6 @@ func (c *Channel) UnmarshalJSON(b []byte) error {
 	ct, ok := raw["type"]
 	if ok {
 		c.Type = ct.(string)
-		delete(raw, "type")
 		c.Raw = raw
 	}
 
@@ -126,16 +125,10 @@ func (c *Channel) UnmarshalJSON(b []byte) error {
 func (c *Channel) MarshalJSON() ([]byte, error) {
 	raw := map[string]interface{}{}
 	if c != nil && c.Type != "" {
-		raw["type"] = c.Type
 		for k, v := range c.Raw {
 			raw[k] = v
 		}
 	}
 
-	b, err := json.Marshal(raw)
-	if err != nil {
-		return nil, err
-	}
-
-	return b, nil
+	return json.Marshal(raw)
 }
--- a/log_entry.go
+++ b/log_entry.go
@@ -40,7 +40,7 @@ type CommonLogEntryField struct {
 // LogEntry is a list of all of the events that happened to an incident.
 type LogEntry struct {
 	CommonLogEntryField
-	Incident Incident
+	Incident Incident `json:"incident"`
 }
 
 // ListLogEntryResponse is the response data when calling the ListLogEntry API endpoint.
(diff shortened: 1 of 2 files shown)


theckman:
So this change is technically breaking API compatibility, and under normal circumstances would be a candidate for a v2.0.0 release. That said, digging further I'm not confident that the old format was actually functional based on the API documentation. As such, I'm not confident we should hold this to the same compatibility guarantee under the v1.0.0 released.

I'm going to sync up with @stmcallister to propose a `v1.5.0` release that includes this change, and other "breaking" changes, which fix an implementation that has never worked reliably.

StupidScience:
@theckman may be better instead of breaking API would be to provide custom `MarshalJSON()` function? It won't break things but should resolve issue as well. WDYT?

theckman:
Yeah, that may be more ideal. Unmarshal into a map, extract the type field out, and then the rest of the map on the Raw field.

theckman (changes requested):
Should have made that comment (^) while requesting changes.

log_entry.go
```diff
@@ -115,8 +115,27 @@ func (c *Channel) UnmarshalJSON(b []byte) error {
 	ct, ok := raw["type"]
 	if ok {
 		c.Type = ct.(string)
+		delete(raw, "type")
```
theckman, line 118: Sorry, I'm going to be a real pain in your rear here... For some reason it didn't dawn on me that removing this would also be a breaking change, and also make it not the raw object anymore. Would you have the spare moment to squash out this line? If not I can fix it after merging this PR.

Thank you so much for this contribution.

log_entry.go
```diff
@@ -115,8 +115,27 @@ func (c *Channel) UnmarshalJSON(b []byte) error {
...
+	if c != nil && c.Type != "" {
+		raw["type"] = c.Type
+		for k, v := range c.Raw {
+			raw[k] = v
+		}
+	}
+
+	b, err := json.Marshal(raw)
+	if err != nil {
+		return nil, err
```
theckman, line 137: I suspect `json.Marshal()` effectively does the same thing. How would you feel about simplifying this error handling block all down to `return json.Marshal(raw)`?

theckman (changes requested):
Requesting two final adjustments, one of them is my fault. Afterwards I think this will be :+1:

log_entry.go
StupidScience: Nice catch. I'll fix it

log_entry.go
StupidScience: Not a problem at all. I'll send update shortly

theckman (approved):

PR #265, opened 2021-02-06
Provide a method for ferrying API errors back to the caller

### LINKED ISSUE #222: Return structured errors from functions that do HTTP requests
This is a similar issue to https://github.com/PagerDuty/go-pagerduty/issues/95, but somewhat distinct.

In `ManageEvent`, several different conditions can return an error. Depending on the type of error delivered, the caller might want to take a variety of actions. However, this is hard-to-impossible currently, since the function just returns a string.

One use case, for instance, is retrying a request if the HTTP status is 5xx, but not retrying the request if the HTTP status is 4xx. Or, truncating the request size if the service returns a 413 error, and trying again.

I'd suggest something like the following for a structured error type:
```
type ManageEventError struct {
  // StatusCode is the status code from the response
  StatusCode int

  // Body is the contents of the response body
  Body []byte

  // Header is the response header
  Header http.Header

  // Message is a human-readable error message
  Message string
}

func (e *ManageEventError) Error() string {
  return e.Message
}
```

Depending on the exact expectations of the submitter of #95, this should be enough information to satisfy their needs as well. The upside of doing this, is that the functionality can be implemented without breaking the API, and so can be done in a minor release. The caller can simply use a type assertion to check further error details.

theckman:
Today the Go PagerDuty API client parses the JSON error response provided by the
API, but does not return it in a way that the caller can use it to make an
informed decision on how to handle that error. For example, being able to
identify if the API request failed or if that resource just wasn't found.

This change adds a new `pagerduty.APIError` type that includes the error object
that gets returned from the API as well as the HTTP Status Code. This should
allow callers to identify if they've ben rate limited, if the resource they were
looking for wasn't found, etc. This new type also includes some helper methods
to further ease the barrier in trying to understand the failure, such as
.RateLimited().

This also adds a short blurb to the `README.md` file to explain how to use this
error type / value.

Fixes #222

### DIFF
--- a/client.go
+++ b/client.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"runtime"
+	"strings"
 	"time"
 )
 
@@ -56,10 +57,73 @@ type APIDetails struct {
 	Details string `json:"details,omitempty"`
 }
 
-type errorObject struct {
-	Code    int         `json:"code,omitempty"`
-	Message string      `json:"message,omitempty"`
-	Errors  interface{} `json:"errors,omitempty"`
+// APIErrorObject represents the object returned by the API when an error
+// occurs. This includes messages that should hopefully provide useful context
+// to the end user.
+type APIErrorObject struct {
+	Code    int      `json:"code,omitempty"`
+	Message string   `json:"message,omitempty"`
+	Errors  []string `json:"errors,omitempty"`
+}
+
+// APIError represents the error response received when an API call fails. The
+// HTTP response code is set inside of the StatusCode field, with the APIError
+// field being the structured JSON error object returned from the API.
+//
+// This type also provides some helper methods like .RateLimited(), .NotFound(),
+// and .Temporary() to help callers reason about how to handle the error.
+//
+// You can read more about the HTTP status codes and API error codes returned
+// from the API here: https://developer.pagerduty.com/docs/rest-api-v2/errors/
+type APIError struct {
+	// StatusCode is the HTTP response status code
+	StatusCode int `json:"-"`
+
+	// APIError represents the object returned by the API when an error occurs.
+	// If the response has no error object present, this will be nil.
+	//
+	// This includes messages that should hopefully provide useful context to
+	// the end user.
+	APIError *APIErrorObject `json:"error"`
+
+	message string
+}
+
+// Error satisfies the error interface, and should contain the StatusCode,
+// APIErrorObject.Message, and APIErrorObject.Code.
+func (a APIError) Error() string {
+	if len(a.message) > 0 {
+		return a.message
+	}
+
+	if a.APIError == nil {
+		return fmt.Sprintf("HTTP response failed with status code %d and no JSON error object was present", a.StatusCode)
+	}
+
+	return fmt.Sprintf(
+		"HTTP response failed with status code %d, message: %s (code: %d)",
+		a.StatusCode, a.APIError.Message, a.APIError.Code,
+	)
+}
+
+// RateLimited returns whether the response had a status of 429, and as such the
+// client is rate limited. The PagerDuty rate limits should reset once per
+// minute, and for the REST API they are an account-wide rate limit (not per
+// API key or IP).
+func (a APIError) RateLimited() bool {
+	return a.StatusCode == http.StatusTooManyRequests
+}
+
+// Temporary returns whether it was a temporary error, one of which is a
+// RateLimited error.
+func (a APIError) Temporary() bool {
+	return a.RateLimited() || (a.StatusCode >= 500 && a.StatusCode < 600)
+}
+
+// NotFound returns whether this was an error where it seems like the resource
+// was not found.
+func (a APIError) NotFound() bool {
+	return a.StatusCode == http.StatusNotFound || a.APIError.Code == 2100
 }
 
 func newDefaultHTTPClient() *http.Client {
@@ -161,7 +225,6 @@ func (c *Client) delete(path string) (*http.Response, error) {
 }
 
 func (c *Client) put(path string, payload interface{}, headers *map[string]string) (*http.Response, error) {
-
 	if payload != nil {
 		data, err := json.Marshal(payload)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


theckman:
@echlebek ping, since you raised the linked issue.

client_test.go
```diff
@@ -40,3 +41,174 @@ func testEqual(t *testing.T, expected interface{}, actual interface{}) {
 		t.Errorf("returned %#v; want %#v", expected, actual)
 	}
 }
+
+func TestAPIError_Error(t *testing.T) {
+	const jsonBody = `{"error":{"code": 420, "message": "Enhance Your Calm", "errors":["Enhance Your Calm", "Slow Your Roll"]}}`
```
echlebek, line 46: :ok_hand: :v:

client.go
```diff
@@ -224,27 +275,33 @@ func (c *Client) checkResponse(resp *http.Response, err error) (*http.Response,
...
-		var eo *errorObject
-		var getErr error
-		if eo, getErr = c.getErrorFromResponse(resp); getErr != nil {
-			return resp, fmt.Errorf("Response did not contain formatted error: %s. HTTP response code: %v. Raw response: %+v", getErr, resp.StatusCode, resp)
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		// check whether the error response is declared as JSON
+		if !strings.HasPrefix(resp.Header.Get("Content-Type"), "application/json") {
+			return resp, fmt.Errorf("HTTP response with status code %d does not contain Content-Type: application/json", resp.StatusCode)
+		}
```
echlebek, line 283: Wondering if there are versions of the API that respond with JSON but don't bother setting this header? Or, are there versions of the API that respond with a 2xx status, but have a JSON-encoded error message?

If yes, is it worth caring about?

echlebek (commented):
Looks good to me, and is a big improvement over the current state of affairs in my opinion!

I do wonder if it's also worth including the response headers in the `APIError type`? I would consider these to be useful for debugging, especially for situations where something is brokering the communication and adding its own headers.

I haven't tested this yet, but if the PR is still open on Monday, I'll try integrating it into the code I was working with when I filed the original issue.

client.go
theckman: In a past life I worked at PagerDuty, and at least when I was there and helped design a version of the API it was important that the `Content-Type` was set and correct. I think we should feel comfortable that will be set when they return JSON from the API (whether success or an error)

I'm expecting their Nginx load balancers may not provide a JSON response if they fail, and so this would not try to parse any sort of junk that may spit back but provide the status code. In that case I'd expect it to be `5XX`, which most consumers would interpret to mean that PagerDuty experienced some sort of failure that resulted in a malformed response. So I think this is okay.

Their documentation indicates that 2XX codes will not use an error JSON document:

> If an API request is successful, PagerDuty will respond to it with HTTP status code 2XX and the resources that match the query.
> 
> If the API request is unsuccessful, PagerDuty will respond with a corresponding HTTP error status code and the response body will contain PagerDuty error details.
- https://developer.pagerduty.com/docs/rest-api-v2/errors/

client_test.go
theckman: 😉

theckman:
@echlebek Regarding the headers, is that something your environment has in-place and if so could you describe it a bit more? I'm genuinely curious.

Sharing this thought so I don't forget it, recognizing this may seem like overkill, but since you can provide your own HTTPClient to the PagerDuty client (`.HTTPClient`), might it make sense to solve that there instead? My assumption is that you'd mostly want to log that error versus do something depending on what it is... but i could be wrong.

theckman:
I'm now realizing I should make `StatusCode == 500` a temporary error too. I'll update the PR real quick.

echlebek:
@theckman perhaps providing a custom HTTPClient does make the most sense for my use case. I can't remember exactly, but I think at the time, I didn't realize the client made that possible.

The software it's been used in on my side is https://github.com/sensu/sensu-pagerduty-handler. It's a small and somewhat roughly-hewn program that's responsible for sending alerts to pagerduty from peoples' Sensu deployments. Typically I don't have visibility into or control over the environment that it's running in.

Something you can probably commiserate with: failure to send a pagerduty alert is a pretty unfortunate circumstance, and people experiencing failures are often looking to diagnose and repair the problem as quickly as possible. So for this use case, more information is almost always better. When the pagerduty client fails, sensu-pagerduty-handler back-propagates a failure event to Sensu (likely ultimately destined for whatever the fallback is when pagerduty is not working). These events are usually quite data-rich and it's here that I'd be wanting to capture as much information as possible.

theckman:
@echlebek I definitely can, and some of this work is related to me building a tool to make it so I can page an entire team if we are having issues getting someone off-hours. 😄 

I also just force-pushed my commit here to update it a little bit. That `APIError` value will now also be returned even if there isn't a JSON error object present, so that you can still inspect the `StatusCode`. I did make the `APIError` field a pointer, so that it could be `nil` in cases where it's not present. I documented the need to do the `nil` check in the type's definition, so that should be present when folks look at documentation / use LSP in their editor.

I think I like where this is at now, especially after refining its behaviors in the case where we have an error without a JSON response.

echlebek:
Thanks so much for your work on this!

client.go
```diff
@@ -56,10 +57,73 @@ type APIDetails struct {
 	Details string `json:"details,omitempty"`
 }
 
-type errorObject struct {
-	Code    int         `json:"code,omitempty"`
-	Message string      `json:"message,omitempty"`
-	Errors  interface{} `json:"errors,omitempty"`
```
stmcallister, line 62: Sorry for my ignorance, but figured I would ask, just to make sure. Wouldn't removing `errorObject` be a breaking change?

stmcallister (commented):
Thanks for adding this @theckman! Looks awesome. I just have one question that I asked inline.

client.go
theckman: @stmcallister only capitalized identifiers are exported (visible) outside of the package, and so `errorObject` could only ever be used by the internals of this project. That should make it totally safe for us to change.

In the interest of complete context sharing... we are _effectively_ renaming this to `APIErrorObject`, and so once we merge / release this any removals or breaking changes to that type would no longer be acceptable because of it being capitalized (exported).

stmcallister: Cool. Thanks for the explanation! This looks great. Going to approve.

stmcallister (approved):
👍 Thank you for improving the Error Responses! 🎉

PR #266, opened 2021-02-06
Prepare internals for exposing context.Context in exported API

theckman:
As of today there is not a way for callers to control the timeout of requests on
a per-call basis, as we do not expose context.Context in the APIs of this
package.

This change updates some of the internal code paths to use context.Context,
making use of a context.TODO() for now (until we can expose context.Context) in
the API.

This change also includes some style changes to help make the code a bit less
dense / more idiomatic in some cases. This also adds a few TODO comments, as a
reminder to come back and look at concerning blocks of code that caught my eye.

### DIFF
--- a/ability.go
+++ b/ability.go
@@ -1,5 +1,7 @@
 package pagerduty
 
+import "context"
+
 // ListAbilityResponse is the response when calling the ListAbility API endpoint.
 type ListAbilityResponse struct {
 	Abilities []string `json:"abilities"`
@@ -7,7 +9,7 @@ type ListAbilityResponse struct {
 
 // ListAbilities lists all abilities on your account.
 func (c *Client) ListAbilities() (*ListAbilityResponse, error) {
-	resp, err := c.get("/abilities")
+	resp, err := c.get(context.TODO(), "/abilities")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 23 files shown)


theckman:
Sorry for this PR being so large. Unfortunately, this one needs to be big because I was changing the method signatures for the things that all the other methods call. So that required touching most files in one-shot to avoid breaking the build.

The ones where we expose this via the exported API can be done in much smaller chunks to make reviewing much easier.

theckman:
I reduced some of the changes I made in this PR since first raising it, and so I'd feel comfortable merging this change without a review.

For context, I don't anticipate that this change is going to be contentious as it's just adding `context.Context` to the call path of the methods that all deal with making HTTP requests. Once this is merged, I'll raise an issue so that we can align on _how_ to expose the `context.Context` to consumers before we start implementing it.

theckman:
Just to be clear, I can't merge without a review but the changes seem innocuous enough that I would if I could.

theckman:
@stmcallister I noticed you didn't :eyes: this PR, and wasn't sure if you have time to review this.

If you don't have time would you be willing to give me a blind-trust Approval so I can merge (I do have merge access, just can't merge my own PRs as we require a review before merging and I'm not a repo admin). 😄

theckman:
Oh, you just did it; heh.

stmcallister (approved):
I like this! Thanks for doing this work! 👍

PR #269, opened 2021-02-12
Update internal HTTP methods to not take pointer to map as argument

theckman:
Map types themselves are reference types, and can be passed in as `nil` even
without declaring the value as a pointer to the map.

Likewise, for maps that are nil the `range` operator is safe and no-ops and so
we do not need a nil check before doing the range. If for some reason we wanted
a check, a length check would be more appropriate than a nil check but it's not
necessary.

### DIFF
--- a/change_events.go
+++ b/change_events.go
@@ -63,7 +63,7 @@ func (c *Client) CreateChangeEvent(e ChangeEvent) (*ChangeEventResponse, error) 
 		changeEventPath,
 		false,
 		bytes.NewBuffer(data),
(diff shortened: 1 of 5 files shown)


theckman:
@stmcallister this is a no-op, so it may make sense to also get it merged before 1.4.0.

stmcallister (approved):
Cool. This looks good. Thanks for the contribution! 🎉

Sjeanpierre (approved):
lgtm

PR #271, opened 2021-02-12
Fix possible nil pointer dereference in APIError.NotFound() method

theckman:
If someone were to receive a PagerDuty response that did not contain the error
JSON object, they would hit a nil pointer dereference if they attempted to call
the .NotFound() method if the status code is not 404.

This fixes it by doing a nil check before attempting to read the value to do the
comparison.


theckman:
I'm also working on an alternative API, which will avoid consumers having the same problem in their code. Since we haven't released v1.4.0 yet we should consider it...

theckman:
I'm going to close this pull request in-favor of #272.

PR #278, opened 2021-02-20
return nil instead of resp on error

### LINKED ISSUE #274: CreateEventWithHTTPClient nil pointer reference
[Line 52 of event.go](https://github.com/PagerDuty/go-pagerduty/blob/master/event.go#L52)

has the following:

```go
	resp, err := client.Do(req)
	if err != nil {
		return &EventResponse{HttpStatus: resp.StatusCode}, err
	}
```

In the case of an actual error, it is very likely that `resp` is nil, and will result in a runtime panic. For example:

```
standby_1  | panic: runtime error: invalid memory address or nil pointer dereference
standby_1  | [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x673028]
standby_1  |
standby_1  | goroutine 1 [running]:
standby_1  | github.com/PagerDuty/go-pagerduty.CreateEventWithHTTPClient(0x7ffe1cd47f17, 0x20, 0x7b3de3, 0x7, 0x7ffe1cd47eed, 0xc, 0xc0004d6340, 0x34, 0x0, 0x0, ...)
standby_1  | 	/Users/tag/pkg/mod/github.com/!pager!duty/go-pagerduty@v1.3.0/event.go:52 +0x588
standby_1  | github.com/PagerDuty/go-pagerduty.CreateEvent(...)
standby_1  | 	/Users/tag/pkg/mod/github.com/!pager!duty/go-pagerduty@v1.3.0/event.go:36
standby_1  | main.notifyPagerduty(0xc0004bfa00, 0x7bdff3, 0x20, 0x7ffe1cd47eed, 0xc, 0x7ffe1cd47f17, 0x20, 0x0, 0x0)
```

The V2 client handles this properly and returns nil. (Also note: I meant to use the V2 API in the first place, so this doesn't actually affect me.)

Thanks for the handy library.

blockpane:
Prevent nil reference on error, resolves #274

### DIFF
--- a/event.go
+++ b/event.go
@@ -49,7 +49,7 @@ func CreateEventWithHTTPClient(e Event, client HTTPClient) (*EventResponse, erro
 	req.Header.Set("Content-Type", "application/json")
 	resp, err := client.Do(req)
 	if err != nil {
-		return &EventResponse{HttpStatus: resp.StatusCode}, err
+		return  nil, fmt.Errorf("failed to action request: %w", err)
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
--- a/event.go
+++ b/event.go
@@ -49,7 +49,7 @@ func CreateEventWithHTTPClient(e Event, client HTTPClient) (*EventResponse, erro
 	req.Header.Set("Content-Type", "application/json")
 	resp, err := client.Do(req)
 	if err != nil {
-		return  nil, fmt.Errorf("failed to action request: %w", err)
+		return nil, fmt.Errorf("failed to action request: %w", err)
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {


theckman (approved):

event.go
```diff
@@ -49,7 +49,7 @@ func CreateEventWithHTTPClient(e Event, client HTTPClient) (*EventResponse, erro
 	req.Header.Set("Content-Type", "application/json")
 	resp, err := client.Do(req)
 	if err != nil {
-		return &EventResponse{HttpStatus: resp.StatusCode}, err
+		return  nil, fmt.Errorf("failed to action request: %w", err)
```
theckman, line 52: Would you be able to `go fmt` this file? Just noticed the whitespace before `nil` when getting ready to merge.

theckman (changes requested):

theckman:
@blockpane wanted to ping again to see if you had a moment to remove the leading whitespace on your change. If not give me a shout.

blockpane:
@theckman Sorry, just caught this request, should be fixed.

theckman (approved):

theckman:
@blockpane thank you for this contribution. It'll probably still be a few days until v1.4.0 is out, and it'll be included there.

If you or anyone else is running into issues with this actively, please give us a shout via a GitHub issue and I can work on backporting just that change as v1.3.1.

Issue #267, opened 2021-02-08
Exposing context.Context in exported API without requiring major version bump


theckman:
~I'm raising this issue in anticipation of #266 being merged without many changes to the implementation.~

**Update:** This proposal has been accepted and will be moving forward.

I'd like to provide consumers the ability to cancel outstanding requests using `context.Context` by creating methods within the package that take it as their first argument. We would not replace the current methods, but instead create a `*WitContext()` variant of each method so that you'd have `GetUser()` and `GetUserWithContext()`. We can accomplish this without code duplication, by having the non-context version call the `*WithContext()` variant and passing in a `context.Background()`. This should make the long-term maintenance minimal.

#### Progress Tracking

- [x] `ability.go` - #283
- [x] `addon.go` - #294
- [x] `business_service.go` - #297 
- [x] `change_events.go` - #284
- [x] `escalation_policy.go` - #303
- [x] `event_v2.go` - #290
- [x] `extension.go` - #301 
- [x] `extension_schema.go` - #289
- [x] `incident.go` - #309 
- [x] `log_entry.go` - #291
- [x] `maintenance_window.go` - #298 
- [x] `notifications.go` - #288
- [x] `on_call.go` - #287
- [x] `priorities.go` - #285
- [x] `ruleset.go` - #306 
- [x] `schedule.go` - #307 
- [x] `service.go` - #299
- [x] `service_dependency.go` - #293 
- [x] `tag.go` - #300
- [x] `team.go` - #308 
- [x] `user.go` - #286
- [x] `vendor.go` - #292

#### New Methods / Functions

For new additions / methods to the package, we SHOULD NOT have a `*WithContext()` variant and instead make the default implementation require a `context.Context`. For v2 I believe the `*WithContext()` variants should be dropped and the `context.Context` moved into the main method.

I think PR #260 is a good example of something we should update to have a `context.Context` as its first argument (no `*WithContext()` variant)

#### Why create such a large API to accomplish this?

Speaking intimately about the origins of this package, it was originally released as `1.0.0` versus a `0.x` version which would have allowed us to more freely explore the API. Because we've already had a v1.0.0 and we use Go Modules, it would be a sizable amount of work for consumers to upgrade to v2+ (even if there aren't many changes they need to make to code).

As such, I think it's prudent for us to provide this functionality without breaking the API so we can better prepare for a 2.0.0 release that fixes any outstanding concerns we have with the API, so the cost that consumers need to pay is worth it.

[...]

PR #289, opened 2021-02-26
Update extension_schema.go to accept a context.Context

theckman:
Updates #267

### DIFF
--- a/extension_schema.go
+++ b/extension_schema.go
@@ -8,6 +8,8 @@ import (
 	"github.com/google/go-querystring/query"
 )
 
+// ExtensionSchema represnts the object presented by the API for each extension
+// schema.
 type ExtensionSchema struct {
 	APIObject
 	IconURL     string   `json:"icon_url"`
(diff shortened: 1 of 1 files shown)


extension_schema.go
```diff
@@ -20,34 +22,57 @@ type ExtensionSchema struct {
...
 type ListExtensionSchemaResponse struct {
 	APIListObject
 	ExtensionSchemas []ExtensionSchema `json:"extension_schemas"`
 }
 
+// ListExtensionSchemaOptions are the options to send with the
+// ListExtensionSchema reques(s).
 type ListExtensionSchemaOptions struct {
 	APIListObject
 	Query string `url:"query,omitempty"`
```
theckman, line 36: @stmcallister I don't see this field documented [here](https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1extension_schemas/get). Does this look like a candidate for removal as part of 1.5, or should this remain?

theckman: Although, now that I look more maybe it's implicitly supported so that you can get the total count as part of the paginated response?

stmcallister: I've reached out to our engineering team to see if we can document the `query` query string parameter.

theckman: @stmcallister since this something we can (and should) solve outside of this PR, are we good to merge this one to call the `context.Context` addition done?

stmcallister (approved):
Looks good! :tad

PR #293, opened 2021-02-26
Update service_dependency.go to accept a context.Context

theckman:
This also updates the new methods to not return the `*http.Response`, as it
doesn't seem needed.

Updates #267

### DIFF
--- a/service_dependency.go
+++ b/service_dependency.go
@@ -25,41 +25,105 @@ type ListServiceDependencies struct {
 }
 
 // ListBusinessServiceDependencies lists dependencies of a business service.
+// It's recommended to use ListBusinessServiceDependenciesWithContext instead.
 func (c *Client) ListBusinessServiceDependencies(businessServiceID string) (*ListServiceDependencies, *http.Response, error) {
-	resp, err := c.get(context.TODO(), "/service_dependencies/business_services/"+businessServiceID)
+	return c.listBusinessServiceDependenciesWithContext(context.Background(), businessServiceID)
+}
+
+// ListBusinessServiceDependenciesWithContext lists dependencies of a business service.
+func (c *Client) ListBusinessServiceDependenciesWithContext(ctx context.Context, businessServiceID string) (*ListServiceDependencies, error) {
+	lsd, _, err := c.listBusinessServiceDependenciesWithContext(ctx, businessServiceID)
+	return lsd, err
+}
+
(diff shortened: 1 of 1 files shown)


service_dependency.go
```diff
@@ -25,41 +25,105 @@ type ListServiceDependencies struct {
...
 
 // ListBusinessServiceDependencies lists dependencies of a business service.
+// It's recommended to use ListBusinessServiceDependenciesWithContext instead.
 func (c *Client) ListBusinessServiceDependencies(businessServiceID string) (*ListServiceDependencies, *http.Response, error) {
-	resp, err := c.get(context.TODO(), "/service_dependencies/business_services/"+businessServiceID)
+	return c.listBusinessServiceDependenciesWithContext(context.Background(), businessServiceID)
+}
+
+// ListBusinessServiceDependenciesWithContext lists dependencies of a business service.
+func (c *Client) ListBusinessServiceDependenciesWithContext(ctx context.Context, businessServiceID string) (*ListServiceDependencies, error) {
```
theckman, line 34: @stmcallister let me know what you think of these methods not including the `*http.Response`. As far as I can tell there isn't a need for the consumer to have the response.

stmcallister: I believe I included `*http.Response` so that folks could get information about the Response, status codes and messages, etc. But, it sounds like the context will provide some of this functionality?

theckman: @stmcallister Not the `context`, I think the new `APIError` type we created will take care of that. Since that should contain all errors in the JSON response, as well as the HTTP status code. If the call returns a `nil` error value, consumers should assume it's a `2xx` response code. If the `error` value is non-nil, then they can assume it was `>299` and inspect the actual error if they really want to know.

The way the `*http.Response` support is implemented today consumers could only get the status code and HTTP headers, but they cannot read the HTTP response body. By default the `http.Response.Body` can only be read once, which we do in `decodeJSON`. Attempts to read it again by callers would result in an `io.EOF` error, and so the `*http.Response` we return today isn't very useful, especially because there don't appear to be any documented HTTP response headers that consumers would care about.

We could fix that bug and make the `*http.Response` useful for consumers, but since there are no documented response headers and we handle the HTTP status code (for failures) in the `APIError` type, I think we should lean-away from returning the `*http.Response` and _potentially_ include its removal in our breaking changes in `v1.5`, since the returned `*http.Response` never fully worked.

stmcallister: Ah, okay. That makes sense. Thanks for the explanation.

stmcallister (approved):
Looks good! 👍

Issue #295, opened 2021-02-27
List* methods should not be paginated by default; create paginated versions


theckman:
In v1 of the API we do not disambiguate between methods that do the pagination for us, and those that do it on our behalf. A good example is `ListBusinessServices()`.

For v2 we should probably update `ListBusinessServices()` to not automatically do pagination, and create a `ListBusinessServicesPaginated()` version instead.

We should probably adopt this naming scheme in v1 for anything that we add that is going to do pagination on the behalf of the user.

theckman:
Another one to fix: https://github.com/PagerDuty/go-pagerduty/blob/77a73c9d16987b91549d60d76d5a075e91400df9/ruleset.go#L142-L154

jparram:
is there any way when using the cli to specify we want more than 25 results?

`./pd service list -team-id TeamId`

andrewwatson:
> is there any way when using the cli to specify we want more than 25 results?
> 
> `./pd service list -team-id TeamId`

No, but I did see another issue asking for that feature to be added....

PR #297, opened 2021-02-27
Update business_service.go to accept a context.Context

theckman:
This also updates the new methods to not return the `*http.Response`, as it
doesn't seem needed.

Updates #267

### DIFF
--- a/business_service.go
+++ b/business_service.go
@@ -47,14 +47,30 @@ type ListBusinessServiceOptions struct {
 	APIListObject
 }
 
-// ListBusinessServices lists existing business services.
+// ListBusinessServices lists existing business services. This method currently
+// handles pagination of the response, so all business services should be
+// present.
+//
+// Please note that the automatic pagination will be removed in v2 of this
+// package, so it's recommended to use ListBusinessServicesPaginated instead.
 func (c *Client) ListBusinessServices(o ListBusinessServiceOptions) (*ListBusinessServicesResponse, error) {
+	bss, err := c.ListBusinessServicesPaginated(context.Background(), o)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ListBusinessServicesResponse{BusinessServices: bss}, nil
+}
+
+// ListBusinessServicesPaginated lists existing business services, automatically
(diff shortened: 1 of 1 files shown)


business_service.go
```diff
@@ -76,55 +87,103 @@ func (c *Client) ListBusinessServices(o ListBusinessServiceOptions) (*ListBusine
...
+// CreateBusinessServiceWithContext instead
 func (c *Client) CreateBusinessService(b *BusinessService) (*BusinessService, *http.Response, error) {
-	data := make(map[string]*BusinessService)
-	data["business_service"] = b
-	resp, err := c.post(context.TODO(), "/business_services", data, nil)
+	return c.createBusinessServiceWithContext(context.Background(), b)
+}
+
+// CreateBusinessServiceWithContext creates a new business service.
+func (c *Client) CreateBusinessServiceWithContext(ctx context.Context, b *BusinessService) (*BusinessService, error) {
```
theckman, line 104: @stmcallister similar to #293, I'm curious of your thoughts around removing the `*http.Response` from these methods.

theckman:
This PR also takes naming inspiration from #295.

business_service.go
```diff
@@ -49,12 +49,23 @@ type ListBusinessServiceOptions struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &ListBusinessServicesResponse{BusinessServices: bss}, nil
+}
+
+// ListBusinessServicesPaginated lists existing business services, automatically
+// handling pagination and returning the full collection.
+func (c *Client) ListBusinessServicesPaginated(ctx context.Context, o ListBusinessServiceOptions) ([]*BusinessService, error) {
```
theckman, line 62: This takes naming inspiration from #295.

stmcallister: Good call.

business_service.go
stmcallister: As mentioned in #293, as long as we can get the same information from somewhere--in this case, from the `APIError` object--we should be okay with this change.

stmcallister (approved):
Thanks for this! 👍

Issue #296, opened 2021-02-27
Methods with implicit pagination should accept an include function


theckman:
Related to #294, in v2 we should investigate having the paginated methods accept a `func(T) bool` where, where `T` is the type of item being paginated and the bool is whether the item should be included. For example, 
```go

func ListUsersPaginated(ctx context.Context, include func(User) bool) ([]User, error) { 
  var collection []User

  // ...

  for {
    if include != nil && include(u) {
      collection = append(collection, u)
    }
  }

  return u, nil
}
```

PR #300, opened 2021-02-27
Update tag.go to accept a context.Context

theckman:
Updates #267

### DIFF
--- a/tag.go
+++ b/tag.go
@@ -57,28 +57,68 @@ type TagAssignment struct {
 	Label string `json:"label,omitempty"`
 }
 
-// ListTags lists tags of your PagerDuty account, optionally filtered by a search query.
+// ListTags lists tags on your PagerDuty account, optionally filtered by a
+// search query. This method currently handles pagination of the response, so
+// all tags matched should be present.
+//
+// Please note that the automatic pagination will be removed in v2 of this
+// package, so it's recommended to use ListTagsPaginated() instead.
 func (c *Client) ListTags(o ListTagOptions) (*ListTagResponse, error) {
-	return getTagList(context.TODO(), c, "", "", o)
+	tags, err := c.ListTagsPaginated(context.Background(), o)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


tag.go
```diff
@@ -57,28 +57,68 @@ type TagAssignment struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &ListTagResponse{Tags: tags}, nil
 }
 
-// CreateTag creates a new tag.
+// ListTagsPaginated lists tags on your PagerDuty account, optionally filtered by a search query.
+func (c *Client) ListTagsPaginated(ctx context.Context, o ListTagOptions) ([]*Tag, error) {
```
theckman, line 76: @stmcallister similar question here as on #297, inspired by #295, as to whether we're okay with this naming convention going forward.

stmcallister: Now that I'm looking at this naming convention for the third time I'm starting to question it. 😬 Wouldn't it make more sense to call these something like `ListAllTags` or `ListTagsAll` because they're gathering up all the pages and presenting a big list? Sorry for not thinking about this earlier.

theckman: @stmcallister I debated this myself for a day or two, and there were a few reasons I ultimately leaned this way:

1. It creates a consistent prefix in the name of the method, which should put them near each other in documentation. Helps with discovery. 
2. Similar to the above, I found it easier to read `<Action><Entity>[<Modifier>]` over `<Action>[<Modifier>]<Entity>`. The optional modifier is at the end and not in the middle.
3. I want to explore #296 as part of v2 as a way for consumers to filter the paginated values, which would make the name `All` not very relevant. It is a 2.0 change so we could rename, but I was thinking less breaking changes the better.

stmcallister: Ah, okay. Your logic makes sense. Let's stick with the `ListTagsPaginated` convention. 👍

stmcallister (approved):
Good stuff!

PR #307, opened 2021-03-06
Update schedule.go to accept a context.Context; fix fatal bug

theckman:
While adding `context.Context` support to the `schedule.go` file, I found a
`nil` map write in PreviewSchedule(). This makes me believe this method may be
unused, otherwise I would expect a panic report.

Updates #267

### DIFF
--- a/schedule.go
+++ b/schedule.go
@@ -68,29 +68,46 @@ type UserReference struct {
 	User APIObject `json:"user"`
 }
 
-// ListSchedules lists the on-call schedules.
+// ListSchedules lists the on-call schedules. It's recommended to use
+// ListSchedulesWithContext instead.
 func (c *Client) ListSchedules(o ListSchedulesOptions) (*ListSchedulesResponse, error) {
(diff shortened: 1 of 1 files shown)


schedule.go
```diff
@@ -101,21 +118,37 @@ type PreviewScheduleOptions struct {
...
+
+// PreviewScheduleWithContext previews what an on-call schedule would look like
+// without saving it.
+func (c *Client) PreviewScheduleWithContext(ctx context.Context, s Schedule, o PreviewScheduleOptions) error {
 	v, err := query.Values(o)
 	if err != nil {
 		return err
 	}
-	var data map[string]Schedule
-	data["schedule"] = s
```
theckman, line 111: These two lines being removed are the bug I am referring to.

stmcallister: Nice catch! I have been able to reproduce the bug you mentioned and verified that this approach no longer panics.

stmcallister (approved):
LGTM! And thanks for the bug fix as well! 🎉 🎉

PR #311, opened 2021-03-11
More analytics

newbootz:
Add aggregated analytics endpoints for Teams and Services

https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1analytics~1metrics~1incidents~1services/post

https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1analytics~1metrics~1incidents~1teams/post

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bin/*
 pd
 vendor/
 .vscode/
+.idea/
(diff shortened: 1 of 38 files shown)


theckman:
@newbootz could you squash this PR into a single commit?

PR #313, opened 2021-03-18
Add Resolution string to ManageIncidentsOptions struct

### LINKED ISSUE #243: ManageIncidents - support set resolution
In case of incident resolution (status is resolved) the [API](https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1incidents/put) allows to set a resolution string.
As far as I can see, this is currently not supported.

theckman:
Fixes #243


stmcallister (approved):
👍

PR #319, opened 2021-03-30
Fix invalid JSON struct tag, and other linter issues

### LINKED ISSUE #317: Fix any linter issues and add missing GoDoc comments
While doing some of the work to add `context.Context` to the repo, I found a few cases where the linters on my system flagged some things to be changed.

I think it makes sense to make this the final change that goes into v1.4.0.

theckman:
This change is a pretty noisy change in that I went through and cleaned up all
of the issues that my linters flagged. I wanted to get this done, so that we can
add them as automatic CI checks for future PRs.

In the process I did find one critical bug, where we had a malformed JSON struct
tag:

https://github.com/PagerDuty/go-pagerduty/blob/69ade4b95ef0ff1d582a7d9f226a98f9db886df1/webhook.go#L44

There are two linter issues we cannot fix until v1.5.0, because they are
breaking changes. As such, the v1.5.0 development line would be the first place
we can enforce those CI checks.

Fixes #317

### DIFF
--- a/ability.go
+++ b/ability.go
@@ -28,12 +28,12 @@ func (c *Client) ListAbilitiesWithContext(ctx context.Context) (*ListAbilityResp
 	return &result, nil
 }
 
-// TestAbility Check if your account has the given ability.
+// TestAbility checks if your account has the given ability.
 func (c *Client) TestAbility(ability string) error {
 	return c.TestAbilityWithContext(context.Background(), ability)
 }
 
-// TestAbility Check if your account has the given ability.
+// TestAbilityWithContext checks if your account has the given ability.
 func (c *Client) TestAbilityWithContext(ctx context.Context, ability string) error {
 	_, err := c.get(ctx, "/abilities/"+ability)
 	return err
--- a/ability_test.go
+++ b/ability_test.go
@@ -12,14 +12,13 @@ func TestAbility_ListAbilities(t *testing.T) {
 
 	mux.HandleFunc("/abilities", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		w.Write([]byte(`{"abilities": ["sso"]}`))
+		_, _ = w.Write([]byte(`{"abilities": ["sso"]}`))
 	})
 
-	var client = &Client{apiEndpoint: server.URL, authToken: "foo", HTTPClient: defaultHTTPClient}
+	client := &Client{apiEndpoint: server.URL, authToken: "foo", HTTPClient: defaultHTTPClient}
 	want := &ListAbilityResponse{Abilities: []string{"sso"}}
 
 	res, err := client.ListAbilities()
-
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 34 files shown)


ability.go
```diff
@@ -28,12 +28,12 @@ func (c *Client) ListAbilitiesWithContext(ctx context.Context) (*ListAbilityResp
...
 }
 
-// TestAbility Check if your account has the given ability.
+// TestAbility checks if your account has the given ability.
 func (c *Client) TestAbility(ability string) error {
 	return c.TestAbilityWithContext(context.Background(), ability)
 }
 
-// TestAbility Check if your account has the given ability.
+// TestAbilityWithContext checks if your account has the given ability.
```
theckman, line 36: This was flagged by `golint`.

ability_test.go
```diff
@@ -12,14 +12,13 @@ func TestAbility_ListAbilities(t *testing.T) {
 
 	mux.HandleFunc("/abilities", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		w.Write([]byte(`{"abilities": ["sso"]}`))
+		_, _ = w.Write([]byte(`{"abilities": ["sso"]}`))
```
theckman, line 15: This was flagged by `errcheck`, which finds places where error checking was missed.

ability_test.go
```diff
@@ -12,14 +12,13 @@ func TestAbility_ListAbilities(t *testing.T) {
...
 	})
 
-	var client = &Client{apiEndpoint: server.URL, authToken: "foo", HTTPClient: defaultHTTPClient}
+	client := &Client{apiEndpoint: server.URL, authToken: "foo", HTTPClient: defaultHTTPClient}
```
theckman, line 18: My editor uses `gofumpt` instead of `go fmt` for style*, and it's a bit more opinionated. This was changed automatically.

* Technically it's `gofumports`, which is meant to replace `goimports`.

analytics_test.go
```diff
@@ -30,14 +30,17 @@ func TestAnalytics_GetAggregatedIncidentData(t *testing.T) {
 	}
 
 	bytesAnalyticsResponse, err := json.Marshal(analyticsResponse)
+	testErrCheck(t, "json.Marshal()", "", err)
```
theckman, line 33: The ineffective assignment linter caught that we weren't handling the `err` value above; this is absolutely a bug.

client.go
```diff
@@ -54,6 +54,8 @@ type APIReference struct {
 	Type string `json:"type,omitempty"`
 }
 
+// APIDetails are the fields required to represent a details non-hydrated
+// object.
```
theckman, line 58: This was caught by `golint`.

analytics_test.go
```diff
@@ -30,14 +30,17 @@ func TestAnalytics_GetAggregatedIncidentData(t *testing.T) {
...
 	mux.HandleFunc("/analytics/metrics/incidents/all", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "POST")
-		w.Write(bytesAnalyticsResponse)
+		_, _ = w.Write(bytesAnalyticsResponse)
 	})
 
-	client := &Client{apiEndpoint: server.URL,
-		authToken:  "foo",
-		HTTPClient: defaultHTTPClient,
+	client := &Client{
```
theckman, line 40: This change was also `gofumports`.

theckman (commented):
I added comments to indicate which linters / formatters caused some changes.

theckman:
@stmcallister wanted to ping again to see if you may have a moment to :eyes: this week. I apologize for it being such a large PR, thankfully the majority of changes were done by automatic tooling and not by hand.

stmcallister (approved):
Looks good! Reading through the lint changes was educational for me. 😁 Also, thanks for catching the webhook bug and bumping the version up to 1.4.0.

PR #322, opened 2021-04-18
Add missing context.Context support and propagation

theckman:
While looking at something I discovered that I had accidentally missed adding
full `context.Context` support in two places. In particular, in
`service_dependency.go` the context for `AssociateServiceDependenciesWithContext`
wasn't being used. Second, I missed adding context support to `AssignTags()`
completely.


theckman:
@stmcallister missed two. :)

stmcallister (approved):
Looks good! 🌮

PR #273, opened 2021-02-15
Adds the ability to define an escalation policy for the manage incidents API

evnsio:

PR #323, opened 2021-04-20
Add ability to set Escalation Policy when managing incidents

theckman:
This was originally suggested by @evnsio in #273, but there were some challenges
getting that PR shephered through and so this is being reintroduced as a new PR.

Replaces #273


theckman:
@stmcallister barring no surprises, this should be the final code change PR. Last PR to come would be the changelog.

stmcallister (approved):
Nice! Thanks for adding `EscalationPolcy`!

Issue #305, opened 2021-03-05
Remove *http.Response returns from API methods


theckman:
Currently we have some methods that return an `*httpResponse`, like this one: https://github.com/PagerDuty/go-pagerduty/blob/1a0c44e17d3ec590336e5d5b256ee123ac6cc023/ruleset.go#L170

Although, digging further I noticed that in many cases we read the response body and don't replace it:

https://github.com/PagerDuty/go-pagerduty/blob/1a0c44e17d3ec590336e5d5b256ee123ac6cc023/ruleset.go#L174

https://github.com/PagerDuty/go-pagerduty/blob/1a0c44e17d3ec590336e5d5b256ee123ac6cc023/ruleset.go#L202

https://github.com/PagerDuty/go-pagerduty/blob/1a0c44e17d3ec590336e5d5b256ee123ac6cc023/client.go#L314-L318

So, the only thing consumers can read from the `*http.Response` are the headers, and not the response. We should just remove this as one of the breaking changes in v1.5.0, since it never really worked.

## Impacted Files
- [x] business_service.go (#358)
- [x] incident.go (#357)
- [x] ruleset.go (#361)
- [x] service_dependency.go (#359)
- [x] tag.go (#360)

## Impacted Methods
* AssignTags
* AssignTagsWithContext
* AssociateServiceDependencies
* CreateBusinessService
* CreateRuleset
* CreateRulesetRule
* CreateRulesetRuleWithContext
* CreateTag
* CreateTagWithContext
* DisassociateServiceDependencies
* DisassociateServiceDependencies
* GetIncidentAlert
* GetRuleset
* GetRulesetRule
* GetTag
* GetTagWithContext
* ListBusinessServiceDependencies
* ListTechnicalServiceDependencies
* ManageIncidentAlerts
* UpdateBusinessService
* UpdateRuleset
* UpdateRulesetRule
* UpdateRulesetRuleWithContext

theckman:
So I think I came up with a pattern we can use that allows us to remove this from the majority of the API, while also giving consumers a debug capability to allow them to capture the response if they need it. I'll raise a PR for that once v1.4.0 is out.

theckman:
So to my comment above, that was #325 which is now merged. I can start looking at this, finally.

theckman:
This is done.

PR #325, opened 2021-04-23
Add support for better API debugging; start v1.5.0 development

theckman:
In prepration for #305, this adds a mechanism to inspect all API requests and
responses handled by the client. That way if you want to figure out why
something isn't working as expected, or you'd like to make use of an
undocumented field/feature, you have access to the response and can use it.

This also adds a `Do()` method, with the same signature as `*http.Client.Do()`,
that allows consumrs to generate their own request, have the client add
authentication details and other headers, before sending the request to the
PagerDuty API. This allows consumers to further debug and make use of features
not yet supported in the REST client.

Also, updates Version string to v1.5.0 as this is work towards that release.

Lastly, this updates some of the test files to remove some duplication across
them. These changes were needed due to the new fields added to the struct that
had to be initialized to make HTTP requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,21 @@ An example of the `service` sub-command
 pd service list
 ```
 
-
 ### Client Library
+
+#### NOTICE: Breaking API Changes in master branch
+
+As part of the upcoming `v1.5.0` release, we will be fixing features that have
+never worked correctly and require a breaking API change to fix. One example is
+the issue reported in [\#232](https://github.com/PagerDuty/go-pagerduty/issues/232),
+as well as a handful of other examples within the
+[v1.5.0 milestone](https://github.com/PagerDuty/go-pagerduty/milestone/2).
+
+As a result, the `master` branch now contains breaking changes since the last
+`v1.4.x` release. We will clearly highlight the breaking changes in the `v1.5.0`
+release notes when it's ready.
+
+#### Example Usage
 
 ```go
 package main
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -42,9 +42,22 @@ An example of the `service` sub-command
 pd service list
 ```
 
-
 ### Client Library
 
+#### NOTICE: Breaking API Changes in master branch
```
theckman, line 47: This isn't technically true yet, but I plan to merge open PRs soon that will make this true... so may as well prematurely optimize. :smile:

change_events_test.go
```diff
@@ -81,12 +76,7 @@ func TestChangeEvent_CreateWithPayloadVerification(t *testing.T) {
 		},
 	)
 
-	client := &Client{
-		v2EventsAPIEndpoint: server.URL,
-		apiEndpoint:         server.URL,
-		authToken:           "foo",
-		HTTPClient:          defaultHTTPClient,
-	}
+	client := defaultTestClient(server.URL, "foo")
```
stmcallister, line 79: Technically, shouldn't these also be passing values for the `v2EventsAPIEndpoint`?

stmcallister (commented):
Overall, looks great! I did have one question on the `change_events_test` that I dropped in that file.

change_events_test.go
theckman: It does: https://github.com/PagerDuty/go-pagerduty/pull/325/files/4f01c5befe6bde8e812f8acf42546764d58bf7c8#diff-4dc56e64f310cd8c12bb00f75c09e83f5d875563ea9901788e525487dc6788b3R41

stmcallister (approved):

PR #326, opened 2021-04-26
Adds v3 webhook signature verification.

ChezCrawford:


dobs (approved):
Only one non-blocking nit, otherwise LGTM!

webhooks.go
```diff
@@ -0,0 +1,62 @@
...
+			continue
+		}
+
+		currentSignatures = append(currentSignatures, currentSignature)
+	}
+
+	return currentSignatures
+}
+
+func calulateSignature(payload []byte, secret string) []byte {
```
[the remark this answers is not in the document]
ChezCrawford: eeee good catch.  Thanks!

webhooks.go
```diff
@@ -0,0 +1,62 @@
...
+var (
+	ErrNoValidSignature = errors.New("invalid webhook signature")
+)
+
+// VerifySignature compares the provided signature of a PagerDuty v3 Webhook
+// against the expected value and returns an error if the values do not match.
+//
+// See https://developer.pagerduty.com/docs/webhooks/webhook-signatures/ for more details.
+//
+func VerifySignature(payload []byte, header, secret string) error {
```
theckman, line 24: My concern with this approach is that implementation details of the PagerDuty API leak outside of the package into the consumer's program, specifically needing to know which header to look for and pull it out.

How would you feel about taking the entire `*http.Request` as the first argument, and then the secret as the second? That way the headers and what-not are not exposed to the consumer. After reading the body we could replace it with a `ioutil.NopCloser(bytes.Reader(body))` so that consumers can then read the request body themselves.

webhooks.go
```diff
@@ -0,0 +1,62 @@
...
+	"strings"
+)
+
+const (
+	currentSignaturePrefix = "v1="
+)
+
+var (
+	ErrNoValidSignature = errors.New("invalid webhook signature")
+)
```
theckman, line 17: Mind collapsing both of these blocks since there is only one item in each.

webhooks.go
```diff
@@ -0,0 +1,62 @@
...
+	"errors"
+	"strings"
+)
+
+const (
+	currentSignaturePrefix = "v1="
+)
+
+var (
+	ErrNoValidSignature = errors.New("invalid webhook signature")
```
theckman, line 16: Can you please add a GoDoc comment here since this is exported?

theckman (changes requested):

webhooks.go
theckman: Thinking more, if we do this we may need another sentinel error for an invalid request (missing / malformed header).

ChezCrawford: Yeah that's not a bad idea.

The intention was definitely to build more helpful public API methods on top of this one.  Was just trying to get some initial code out in public since we have folks asking about how this would be implemented.

ChezCrawford: @theckman : actually I am kind of thinking that the method you are describing might be in addition to this one.  Something like: 

```go
// VerifyWebhookRequest extracts and verifies the webhook signature.
// This method would likely return more detailed errors that differentiate between missing / invalid signatures, etc.
func VerifyWebhookRequest(request *http.Request, secret string) error
```

or
 
```go
// ConstructWebhookEvent verifies and parses a v3 Webhook Event
func ConstructWebhookEvent(request *http.Request, secret string) WebhookEvent, error
```

It still seems useful leave this more raw public method for clients wishing to have more control over this functionality?

theckman: My personal preference _generally_ is to try and keep the API as small as we can, as it means there's less for us to maintain and honestly less to potentially break by accident in a future update. Text communication is hard, so maybe it'd be best if I try to figure out what assumptions / expectations I think I have:

* I believe the majority of Gophers are working within an environment where the request will be available to them as an `*http.Request`.
* Once it's out of early access, I don't expect the V3 signature format to change much (especially the header) without it incurring a new version designation from PagerDuty.
* Considering it's unlikely to change, someone could lift the verification code out of this project and put it into theirs (or be inspired by it to write their own) if our API signature doesn't work for them.

If we get issues + comments indicating that the `*http.Request` format doesn't work for many, we can look to expose that additional functionality in a future minor release. If we make the API too wide right away, we must release a major version if we realize nobody is using it and no longer wish to support it.

PR #329, opened 2021-05-10
Fix query params on tags

jfmyers9:
At the moment, this method is missing the `?` which results in a 404 when providing query parameters.


theckman (approved):
Nice find; this also made me realize there is another case of this elsewhere that I will raise a PR to go with this one into v1.4.1.

theckman:
@jfmyers9 Thanks again for this PR! Just released `v1.4.1` which includes this fix.

PR #330, opened 2021-05-11
Fix pagination within ListBusinessServices + Paginated

theckman:
When reviewing #329, I found that `ListBusinessServicesPaginated` and by proxy
`ListBusinessServices`, were not properly encoding their URL parameters,
effectively breaking pagination due to the missing ? before the values.

This adds the missing ? to the URL.


stmcallister (approved):
Good catch! Sorry I missed this earlier. 🤦

PR #338, opened 2021-06-23
[Incident][Alert] Add From Parameter

tuannvm:
## Why?

- Add `from` parameter
- According to [API document](https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1incidents~1%7Bid%7D~1alerts/put), `from` header is mandatory

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -625,18 +625,23 @@ func (c *Client) getIncidentAlertWithContext(ctx context.Context, incidentID, al
 
 // ManageIncidentAlerts allows you to manage the alerts of an incident. It's
 // recommended to use ManageIncidentAlertsWithContext instead.
-func (c *Client) ManageIncidentAlerts(incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, *http.Response, error) {
-	return c.manageIncidentAlertsWithContext(context.Background(), incidentID, alerts)
+func (c *Client) ManageIncidentAlerts(from string, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, *http.Response, error) {
+	return c.manageIncidentAlertsWithContext(context.Background(), from, incidentID, alerts)
 }
 
 // ManageIncidentAlertsWithContext allows you to manage the alerts of an incident.
-func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, error) {
-	lar, _, err := c.manageIncidentAlertsWithContext(context.Background(), incidentID, alerts)
+func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, from string, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, error) {
+	lar, _, err := c.manageIncidentAlertsWithContext(context.Background(), from, incidentID, alerts)
 	return lar, err
 }
 
-func (c *Client) manageIncidentAlertsWithContext(ctx context.Context, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, *http.Response, error) {
-	resp, err := c.put(ctx, "/incidents/"+incidentID+"/alerts/", alerts, nil)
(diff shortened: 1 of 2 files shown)


incident.go
```diff
@@ -625,18 +625,23 @@ func (c *Client) getIncidentAlertWithContext(ctx context.Context, incidentID, al
...
-func (c *Client) ManageIncidentAlerts(incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, *http.Response, error) {
-	return c.manageIncidentAlertsWithContext(context.Background(), incidentID, alerts)
+func (c *Client) ManageIncidentAlerts(from string, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, *http.Response, error) {
+	return c.manageIncidentAlertsWithContext(context.Background(), from, incidentID, alerts)
 }
 
 // ManageIncidentAlertsWithContext allows you to manage the alerts of an incident.
-func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, error) {
-	lar, _, err := c.manageIncidentAlertsWithContext(context.Background(), incidentID, alerts)
+func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, from string, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, error) {
```
theckman, line 633: Could you add to the GoDoc comment of both of these methods some context around the `from` parameter? Maybe something like `// The from parameter is the email address of a valid user, associated with the PagerDuty account, making the request.`

incident.go
```diff
@@ -625,18 +625,23 @@ func (c *Client) getIncidentAlertWithContext(ctx context.Context, incidentID, al
...
+	lar, _, err := c.manageIncidentAlertsWithContext(context.Background(), from, incidentID, alerts)
 	return lar, err
 }
 
-func (c *Client) manageIncidentAlertsWithContext(ctx context.Context, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, *http.Response, error) {
-	resp, err := c.put(ctx, "/incidents/"+incidentID+"/alerts/", alerts, nil)
+func (c *Client) manageIncidentAlertsWithContext(ctx context.Context, from string, incidentID string, alerts *IncidentAlertList) (*ListAlertsResponse, *http.Response, error) {
+
```
theckman, line 639: Could you remove this line of whitespace from the top of the function, so `h := ...` is right up against the header?

incident_test.go
```diff
@@ -649,7 +649,7 @@ func TestIncident_ManageAlerts(t *testing.T) {
 			},
 		},
 	}
-	res, _, err := client.ManageIncidentAlerts(incidentID, input)
+	res, _, err := client.ManageIncidentAlerts(from, incidentID, input)
```
theckman, line 652: This doesn't seem to actually assert that the `from` value is making it through as an HTTP header. Could you update the handler above (on line `635`) to look for that header and the correct value, and fail the request/test otherwise?

incident.go
```diff
@@ -625,18 +625,23 @@ func (c *Client) getIncidentAlertWithContext(ctx context.Context, incidentID, al
...
+	h := map[string]string{
+		"From": from,
+	}
+
+	resp, err := c.put(ctx, "/incidents/"+incidentID+"/alerts/", alerts, h)
```
theckman, line 644: I know this existed before you made the change, but it seems like the trailing `/` on `/alerts/` is a mistake. Could you remove it so that it just reads `"/alerts"`?

theckman (changes requested):

theckman:
Thank you for the contribution! There are a few opportunities to improve this PR, and so I added some comments with my feedback/thoughts. Let me know if you have any questions.

theckman:
@tuannvm thank you again for this contribution. I wanted to reach back out to see if you had a moment to address the comments I made above, as well as whether you can resolve the merge conflicts and then rebase/squash your commit(s)? The `v1.5.0` merge window is now open and I'd like to make sure this is included.

theckman:
I opened #380 to supersede this PR.

PR #340, opened 2021-06-30
Correct formatting of deprecation notices.

dsymonds:
There is a specific format for deprecation notices in Go doc comments
(see https://golang.org/wiki/Deprecated) that ensures they are correctly
rendered (e.g.
https://pkg.go.dev/github.com/PagerDuty/go-pagerduty#Client.CreateIncidentNote).

Add a handful of missing deprecation notices to functions that now have
parallel WithContext counterparts.

### DIFF
--- a/ability.go
+++ b/ability.go
@@ -7,8 +7,9 @@ type ListAbilityResponse struct {
 	Abilities []string `json:"abilities"`
 }
 
-// ListAbilities lists all abilities on your account. It's recommended to use
-// ListAbilitiesWithContext instead.
+// ListAbilities lists all abilities on your account.
+//
+// Deprecated: Use ListAbilitiesWithContext instead.
 func (c *Client) ListAbilities() (*ListAbilityResponse, error) {
 	return c.ListAbilitiesWithContext(context.Background())
 }
@@ -29,6 +30,8 @@ func (c *Client) ListAbilitiesWithContext(ctx context.Context) (*ListAbilityResp
 }
 
 // TestAbility checks if your account has the given ability.
+//
+// Deprecated: Use TestAbilityWithContext instead.
 func (c *Client) TestAbility(ability string) error {
 	return c.TestAbilityWithContext(context.Background(), ability)
 }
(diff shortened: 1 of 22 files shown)


theckman:
I was sorta reluctant to mark these as deprecated in this way, as the methods aren't going away in v2. Instead, the WithContext ones will be renamed effectively changing the parameters of the current ones.

dsymonds:
Well, some are already marked as deprecated, just inconsistently, and even until v2 it seems good to steer people away from the non-context versions. More importantly, even when v2 comes along, v1 will be around for a long time, so if you're using it you need to know about this.

ability.go
```diff
@@ -7,8 +7,8 @@ type ListAbilityResponse struct {
 	Abilities []string `json:"abilities"`
 }
 
-// ListAbilities lists all abilities on your account. It's recommended to use
-// ListAbilitiesWithContext instead.
+// ListAbilities lists all abilities on your account.
+// Deprecated: Use ListAbilitiesWithContext instead.
```
theckman, line 11: @dsymonds I took a look at the docs on this comment, and it seems like it expects the deprecation notice to be on its own paragraph: https://github.com/golang/go/wiki/Deprecated

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with Deprecated: followed by some information about the deprecation, and a recommendation on what to use instead, if applicable.
> 
> The paragraph does not have to be the last paragraph in the doc comment.

Would you be able to update the change to look more like:

```go
// ListAbilities lists all abilities on your account.
//
// Deprecated: Use ListAbilitiesWithContext instead.
```

theckman (changes requested):
Looking to get this in as part of the `v1.5.0` release, but I think the comments need to be updated based on the context in the wiki. I've added a single comment in my review, but it applies to all of the proposed changes I believe.

ability.go
dsymonds: Sure, done.

theckman (approved):
I'll merge next time I'm in front of my workstation. Thank you!

theckman:
Finally got my pending PR approved by a PD employee and so I could merge this. Thanks again for the contribution!

PR #345, opened 2021-08-09
allow setting suppress to false

cluarkhpe:
The Pagerduty API allows setting suppress to false. omitempty prevents this from working.


theckman (approved):
Thank you for the contribution.

PR #347, opened 2021-08-13
Adding support for analytics raw data - multiple incidents

jozephp:
Support for Analytics -Get aggregated incident data 
 https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1analytics~1metrics~1incidents~1all/post

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -2,6 +2,7 @@ package pagerduty
 
 import (
 	"context"
+	"fmt"
 )
 
 // AnalyticsRequest represents the request to be sent to PagerDuty when you want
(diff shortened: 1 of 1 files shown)


analytics.go
```diff
@@ -58,6 +59,58 @@ type AnalyticsData struct {
...
+	Limit    uint                       `json:"limit,omitempty"`
+	More     bool                       `url:"more,omitempty"`
+	Order    string                     `json:"order,omitempty"`
+	OrderBy  string                     `json:"order_by,omitempty"`
+	Filter   *AnalyticsFilter           `json:"filters,omitempty"`
+	TimeZone string                     `json:"time_zone,omitempty"`
+	Data     []AnalyticsRawIncidentData `json:"data,omitempty"`
+}
+
+type IncidentAnalyticsRequest struct {
```
theckman, line 104: This name looks different than the rest, as it starts with `Incident` instead of `Analytics`. Should this instead be `AnalyticsRawIncidentsRequest`? Plural is intentional on this one too since it could contain the data for multiple incidents.

analytics.go
```diff
@@ -113,3 +166,28 @@ func (c *Client) GetAggregatedTeamData(ctx context.Context, analytics AnalyticsR
 
 	return analyticsResponse, nil
 }
+
+// Get raw data multiple incidents gets the incidents and its metrics for the requested data.
+func (c *Client) GetMultipleRawIncidents(o IncidentAnalyticsRequest) (AnalyticsIncidentResponse, error) {
+	return c.GetMultipleRawIncidentsWithContext(context.Background(), o)
+}
+
+// Gets raw data multiple incidents with context gets the incidents and its metrics for the requested data.
```
theckman, line 175: This GoDoc comment is not formatted correctly. It should start with the method name, and then explain what it is / does.

In addition, could you add a separate newline separated section below indicating it's also using an early access API, and could be subject to change? We probably can't explicitly guarantee API compatibility until the REST API itself is stable.

So, including other suggested changes, it would be something like:

```Go
// GetRawIncidentsAnalytics gets the raw analytics for multiple incidents based on
// the filter provided within the AnalyticsRawIncidentsRequest.
//
// This method currently uses an Early Access REST API endpoint, and as a result
// may be subject to change until the API endpoint becomes stable.
```

theckman (changes requested):
Thank you for your contribution! There are a few comments I added with some requested changes before we'll merge this in to be part of the `v1.5.0` release.

Also, could you please implement a unit tests that make sure the `X-EARLY-ACCESS` header is set on the request, and that JSON deserialization seems to work okay with our defined types?

theckman:
@jozephp might you have some time to make the requested changes to this PR? It's something I'm hoping to get in v1.5.0.

theckman:
@jozephp thought I'd ping one more time to see if you're able to make the requested changes.

theckman:
Just noticed this is using undocumented API fields. I've pinged PagerDuty support to ask about them as we don't tend to support undocumented fields within this library.

ChuckCrawford:
Should be resolved under https://github.com/PagerDuty/go-pagerduty/pull/481

PR #350, opened 2021-08-29
Swap two transposed words in the README file

theckman:


stmcallister (approved):
Maybe we worded it like this on purpose.... 🤦

theckman:
Oh no, it was 100% me. We both missed it on that v1.5.0 PR that was roughly the size of Texas. 😅

PR #359, opened 2021-09-04
Remove returned *http.Response from svc dependency related methods

theckman:
This is a breaking change.

This is one of the changes that will need to be made to complete issue #305,
which is being done in a minor release even though it does contain breaking
changes.

The internal implementation details of how this was implemented meant it never
worked in a way that consumers could use, because the body was always empty.
Seeing as this API never worked and we do not wish to support it, we are going
to remove it so that anyone who may have depended on it, but missed that it was
broken, can be made aware.

If someone wanted to capture the full response of these API calls, they can now
do so using the the functionality added in #325 (4f01c5befe6bde8e812f8acf42546764d58bf7c8).

### DIFF
--- a/service_dependency.go
+++ b/service_dependency.go
@@ -2,7 +2,6 @@ package pagerduty
 
 import (
 	"context"
-	"net/http"
 )
 
 // ServiceDependency represents a relationship between a business and technical service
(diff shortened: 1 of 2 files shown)


stmcallister (approved):
Looks consistent with the other related PRs. Thank you!!! 🎉

PR #363, opened 2021-09-13
implement missing maintenance-window subcommands

Hsn723:
The `pd maintenance-window` command was missing subcommands other than `create`, so I went ahead and implemented them. `create` is also using a deprecated library function so I also addressed that.

### DIFF
--- a/command/maintenance_window_create.go
+++ b/command/maintenance_window_create.go
@@ -1,13 +1,15 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/PagerDuty/go-pagerduty"
-	log "github.com/sirupsen/logrus"
-	"github.com/mitchellh/cli"
 	"os"
 	"strings"
+
+	"github.com/PagerDuty/go-pagerduty"
+	"github.com/mitchellh/cli"
+	log "github.com/sirupsen/logrus"
 )
 
 type MaintenanceWindowCreate struct {
@@ -59,7 +61,7 @@ func (c *MaintenanceWindowCreate) Run(args []string) int {
 		return -1
 	}
 	log.Debugf("%#v", m)
-	if _, err := client.CreateMaintenanceWindows(m); err != nil {
+	if _, err := client.CreateMaintenanceWindowWithContext(context.Background(), "", m); err != nil {
 		log.Error(err)
 		return -1
 	}
--- a/command/maintenance_window_create.go
+++ b/command/maintenance_window_create.go
@@ -1,15 +1,13 @@
 package main
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/PagerDuty/go-pagerduty"
+	log "github.com/sirupsen/logrus"
+	"github.com/mitchellh/cli"
 	"os"
 	"strings"
-
-	"github.com/PagerDuty/go-pagerduty"
-	"github.com/mitchellh/cli"
-	log "github.com/sirupsen/logrus"
 )
 
 type MaintenanceWindowCreate struct {
@@ -61,7 +59,7 @@ func (c *MaintenanceWindowCreate) Run(args []string) int {
 		return -1
 	}
 	log.Debugf("%#v", m)
-	if _, err := client.CreateMaintenanceWindowWithContext(context.Background(), "", m); err != nil {
+	if _, err := client.CreateMaintenanceWindows(m); err != nil {
 		log.Error(err)
 		return -1
 	}
--- a/command/maintenance_window_delete.go
+++ b/command/maintenance_window_delete.go
@@ -1,11 +1,16 @@
 package main
 
 import (
-	"github.com/mitchellh/cli"
+	"context"
+	"fmt"
 	"strings"
+
+	"github.com/mitchellh/cli"
+	log "github.com/sirupsen/logrus"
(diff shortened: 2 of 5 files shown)


command/maintenance_window_delete.go
```diff
@@ -23,5 +32,27 @@ func (c *MaintenanceWindowDelete) Synopsis() string {
 }
 
 func (c *MaintenanceWindowDelete) Run(args []string) int {
+	var mwID string
+	flags := c.Meta.FlagSet("maintenance-window delete")
+	flags.Usage = func() { fmt.Println(c.Help())}
```
theckman, line 37: Based on how this line is formatted, it seems like `go fmt` wasn't ran on save. Could you please run `go fmt` on the files changed in this PR?

command/maintenance_window_delete.go
```diff
@@ -23,5 +32,27 @@ func (c *MaintenanceWindowDelete) Synopsis() string {
...
+
+	if err := flags.Parse(args); err != nil {
+		log.Error(err)
+		return -1
+	}
+	if err := c.Meta.Setup(); err != nil {
+		log.Error(err)
+		return -1
+	}
+	if mwID == "" {
```
theckman, line 48: Does it make sense to do this check before `c.Meta.Setup()`?

command/maintenance_window_list.go
```diff
@@ -23,5 +38,46 @@ func (c *MaintenanceWindowList) Synopsis() string {
 }
 
 func (c *MaintenanceWindowList) Run(args []string) int {
+	var includes, serviceIDs, teamIDs []string
+	var query, filter string
+	flags := c.Meta.FlagSet("maintenance-window list")
+	flags.Usage = func() { fmt.Println(c.Help())}
```
theckman, line 44: This also looks like `go fmt` wasn't ran on save, and so do the rest of the files.

command/maintenance_window_delete.go
```diff
@@ -23,5 +32,27 @@ func (c *MaintenanceWindowDelete) Synopsis() string {
...
+		log.Error("You must provide a maintenance window ID")
+		return -1
+	}
+	client := c.Meta.Client()
+	if err := client.DeleteMaintenanceWindowWithContext(context.Background(), mwID); err != nil {
+		log.Error(err)
+		return -1
+	}
```
theckman, line 56: Slight style nit, could you space these blocks out into logical groupings with a newline in between?

```go
	if err := c.Meta.Setup(); err != nil {
		log.Error(err)
		return -1
	}

	if mwID == "" {
		log.Error("You must provide a maintenance window ID")
		return -1
	}

	client := c.Meta.Client()
	if err := client.DeleteMaintenanceWindowWithContext(context.Background(), mwID); err != nil {
		log.Error(err)
		return -1
	}

	return 0
```

command/maintenance_window_list.go
```diff
@@ -23,5 +38,46 @@ func (c *MaintenanceWindowList) Synopsis() string {
...
+	}
+	for i, mw := range mws.MaintenanceWindows {
+		fmt.Println("Entry: ", i)
+		data, err := yaml.Marshal(mw)
+		if err != nil {
+			log.Error(err)
+			return -1
+		}
+		fmt.Println(string(data))
+	}
```
theckman, line 80: Same comment on the style nit around this block of code and the logical groups. It seems like it also applies to the rest of the files.

command/maintenance_window_create.go
```diff
@@ -59,7 +61,7 @@ func (c *MaintenanceWindowCreate) Run(args []string) int {
 		return -1
 	}
 	log.Debugf("%#v", m)
-	if _, err := client.CreateMaintenanceWindows(m); err != nil {
+	if _, err := client.CreateMaintenanceWindowWithContext(context.Background(), "", m); err != nil {
```
theckman, line 64: Could you undo this change? Since we're not actually making use of the `context` value here, there isn't a reason to update this to call the new method. I'm okay with the new code making use of the new methods. :+1:

command/maintenance_window_update.go
```diff
@@ -23,5 +33,38 @@ func (c *MaintenanceWindowUpdate) Synopsis() string {
...
+	}
+	log.Info("Input file is:", flags.Arg(0))
+	f, err := os.Open(flags.Arg(0))
+	if err != nil {
+		log.Error(err)
+		return -1
+	}
+	defer f.Close()
+	decoder := json.NewDecoder(f)
+	if err := decoder.Decode(&mw); err != nil {
```
theckman, line 60: Since this is a single JSON document and not a stream of them, could we switch this to use `ioutil.ReadFile` + `json.Unmarshal`? There are certain cases where the `json.Decoder` fails to detect malformed JSON, so there are reasons to avoid it unless you're streaming multiple JSON documents.

theckman (changes requested):
Thank you for the contribution. I did request a few changes.

Hsn723:
@theckman 
Thank you for the review. I have applied the requested changes. Please take a look

theckman (approved):
Thank you for your contribution, as well as working with me on the requested changes. It'll probably still take a little bit of time until the `v1.5.0` tag is cut on the repo, but within the next few minutes your changes should be available on the `master` branch if you'd like to start making use of them.

PR #365, opened 2021-09-23
Add json field incidents_responders to Incident struct

sostakas:
Add json:"incidents_responders,omitempty" to Incident struct in order to get
"incident": {"incidents_responders": []} from func GetIncident(ctx context.Context, client *pagerduty.Client, id string) (*pagerduty.Incident, error)

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -89,6 +89,7 @@ type Incident struct {
 	Body                 IncidentBody         `json:"body,omitempty"`
 	IsMergeable          bool                 `json:"is_mergeable,omitempty"`
 	ConferenceBridge     *ConferenceBridge    `json:"conference_bridge,omitempty"`
+	IncidentResponders   []IncidentResponders `json:"incidents_responders,omitempty"`
 }
 
 // ListIncidentsResponse is the response structure when calling the ListIncident API endpoint.
--- a/incident_test.go
+++ b/incident_test.go
@@ -237,7 +237,7 @@ func TestIncident_Get(t *testing.T) {
 
 	mux.HandleFunc("/incidents/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		_, _ = w.Write([]byte(`{"incident": {"id": "1"}}`))
+		_, _ = w.Write([]byte(`{"incident": {"id": "1", "incidents_responders": [{"incident": {"id": "1"}}]}}`))
 	})
 
 	client := defaultTestClient(server.URL, "foo")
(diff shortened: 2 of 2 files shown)


GiedriusS (approved):

theckman:
@sostakas according to the PagerDuty API documentation, that field is not present when getting an incident: https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1incidents~1%7Bid%7D/get

Could you point me to the documentation where it shows that field being returned?

incident.go
```diff
@@ -89,6 +89,7 @@ type Incident struct {
 	Body                 IncidentBody         `json:"body,omitempty"`
 	IsMergeable          bool                 `json:"is_mergeable,omitempty"`
 	ConferenceBridge     *ConferenceBridge    `json:"conference_bridge,omitempty"`
+	IncidentResponders   []IncidentResponders `json:"incidents_responders,omitempty"`
```
theckman, line 92: Could you implement a test that ensures this field is being passed along in the JSON within the request body?

theckman (changes requested):
@sostakas Thank you for the contribution. I did add one comment on the PR requesting a change.

Secondarily, according to the PagerDuty API documentation that field is not present when getting an incident: https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1incidents~1%7Bid%7D/get

Could you point me to the documentation where it shows that field being returned? We tend to only support fields that are explicitly documented by PagerDuty.

incident.go
sostakas: @theckman Thanks for comment. I added the test. It is not documented but you can check response body using `Try it`. This field is must to test if additional responders were added successfully (if you want to stick to go-pagerduty and do not want to create your own http requests)

theckman: Thank you for that context. The policy set by PagerDuty on this library is that only fields that are explicitly listed in the documentation are merged in.

I'll label this PR as being focused on an undocumented feature, and try to engage with PagerDuty to see if they'd be willing/able to document the field(s). I'm not an employee, so I'll probably end up filing a support ticket to make the request to have the documentation updated.

incident_test.go
```diff
@@ -237,15 +237,15 @@ func TestIncident_Get(t *testing.T) {
 
 	mux.HandleFunc("/incidents/1", func(w http.ResponseWriter, r *http.Request) {
 		testMethod(t, r, "GET")
-		_, _ = w.Write([]byte(`{"incident": {"id": "1"}}`))
+		_, _ = w.Write([]byte(`{"incident": {"id": "1", "incidents_responders": [{"incident": {"id": "1"}}]}}`))
```
theckman, line 240: Thank you for doing this. 👍

incident.go
sostakas: @theckman Thank you!

incident_test.go
sostakas: @theckman 👍

theckman (approved):

PR #366, opened 2021-09-27
Adding escalation level to ManageIncidentOptions

sim1s:


theckman (changes requested):
Thank you for the contribution. I have one observation / question about the tests, but beyond that this looks good.

theckman:
@sim1s wanted to see if you'd have some time to take a look at the question I had in the requested changes.

theckman (approved):
I'll address my comments with the tests at a later time. This is true not only of these tests but others.

PR #368, opened 2021-10-06
Return EventsAPIV2Error from Events V2 API call.

abhishekrb19:
This patch does the following:
- Return `EventsAPIV2Error` instead of `error` for the Events V2 API call. This will be consistent with the other REST API calls and callers could leverage the wrapper methods added such as `.BadRequest()`, `.RateLimited()`,
and `.Temporary()`. Add unit tests for the same.
- Note that this only works for Events V2 API call. Specifically, Events V1 has a different error structure and slightly different error code. Hence, the naming to avoid disambiguation.
- Fix a few typos in the code base in client.go.
- Add compile-time type assertions for leveraged built-in interfaces.

### DIFF
--- a/client.go
+++ b/client.go
@@ -169,7 +169,7 @@ func (a APIError) Error() string {
 func apiErrorsDetailString(errs []string) string {
 	switch n := len(errs); n {
 	case 0:
-		panic("errs slice is empty")
+		return ""
 
 	case 1:
 		return errs[0]
--- a/client.go
+++ b/client.go
@@ -90,7 +90,7 @@ type APIErrorObject struct {
 // While the PagerDuty REST API is documented to always return the error object,
 // we assume it's possible in exceptional failure modes for this to be omitted.
 // As such, this wrapper type provides us a way to check if the object was
-// provided while avoiding consnumers accidentally missing a nil pointer check,
+// provided while avoiding consumers accidentally missing a nil pointer check,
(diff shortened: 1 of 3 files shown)


theckman:
@abhishekrb19 might you have some time soon to look at the requested changes?

abhishekrb19:
@theckman, thank you for the review and reminder.
You're right - I missed the documentation of `APIError` - that it was designed to work only with the REST API. Also, the error response is different for the EventV2 API and returns only a small number of response codes - https://developer.pagerduty.com/api-reference/b3A6Mjc0ODI2Nw-send-an-event-to-pager-duty.

I adapted your recommendation to use `EventsAPIV2Error` instead of `EventsAPIError` to avoid ambiguity since there is also an Events V1 API where the structure of JSON response is different.

theckman (changes requested):
I've noticed a few things that should probably be adjusted before merging this PR.

I did also notice that the `ManageEvent*` method isn't included in this change, but it will need to be. I think to do this we'll need to create separate `checkResponse` and `getErrorFromResponse` methods for the Events V2 API calls within the `client.go` file.

theckman:
@abhishekrb19 might you have the time to make the requested changes, so that this can make it into v1.5.0?

abhishekrb19:
@theckman, sorry for not getting back earlier! Thank you for addressing the comments!

Issue #367, opened 2021-10-05
Webhooks V3 Support


thde:
As V1/V2 webhooks are being deprecated, are there plans to support V3 webhooks within `go-pagerduty`?

> We wanted to let you know that Webhooks V3 is now live and available to everyone. Thank you for your continued feedback and support. Learn more about it "here":https://support.pagerduty.com/docs/webhooks.
> This also means that Webhooks V1 and V2 are being deprecated on November 13, 2021 and March 31, 2022 respectively.

I've seen that #326 adds signature verification and should land in v1.5.0?
But the payload also seems to change between [V2](https://developer.pagerduty.com/docs/webhooks/v2-overview/#webhook-payload) and [V3](https://developer.pagerduty.com/docs/webhooks/v2-overview/#webhook-payload).

theckman:
@ChezCrawford I've managed to get everything else for `v1.5.0` ready go. Have you had the cycles to look at this at all? If not no sweat, and I'll retag it to be part of `v1.6.0`.

ChezCrawford:
It's in the works but isn't quite ready for use. I'll shoot for something in v1.6.0.

s-buhar0v:
> This also means that Webhooks V1 and V2 are being deprecated on November 13, 2021 and March 31, 2022 respectively.
Guys, any updates? @ChezCrawford

ChezCrawford:
Alright two threads here: 

First, i've got a draft of some basic support for V3 Webhook Payloads in https://github.com/PagerDuty/go-pagerduty/pull/427 and would appreciate and/all feedback.

Second: My apologies for the delay in communication regarding deprecation; this is has been a bit of a mess from our end. What @theckman said above is essentially correct. Furthermore: be on the lookout for some updated communication regarding webhook deprecation. We are going to do our best not to leave anyone out in the cold 😉 .

PR #370, opened 2021-10-10
Add v3 webhook signature verification

theckman:
This change is an extension of the PR raised by @ChezCrawford (#326), where the
requested changes have been applied to the branch and all commits have been
squashed.

This adds a new function to the package, VerifySignatureWebhookV3, which accepts
an *http.Request and a secret, and validates that the request is properly signed
using that secret. This function does return some sentinel error values so that
you can choose which HTTP status to send back to the caller.

Supersedes #326

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,14 @@ if ok { // resp is an *http.Response we can inspect
 }
 ```
 
+#### Included Packages
+
+##### webhookv3
+
+Support for V3 of PagerDuty Webhooks is provided via the `webhookv3` package.
+The intent is for this package to provide signature verification and decoding
+helpers.
+
 ## Contributing
 
 1. Fork it ( https://github.com/PagerDuty/go-pagerduty/fork )
--- a/examples/webhooks/webhook_server.go
+++ b/examples/webhooks/webhook_server.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/PagerDuty/go-pagerduty/webhookv3"
+)
+
+const (
+	secret = "lDQHScfUeXUKaQRNF+8XIiDKZ7XX3itBAYzwU0TARw8lJqRnkKl2iB1anSb0Z+IK"
+)
+
+func main() {
+	http.HandleFunc("/webhook", handler)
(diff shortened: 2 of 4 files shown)


theckman:
I do what to think on this a little bit. It may make sense to put the Webhook related code within its own package, `webhookv3` for example, considering that we may want to implement struct types to make it easier for folks to parse webhook requests.

theckman:
@ChezCrawford In general I try to qualify the validity of requests before I hand them to my actual handlers. So I'd probably verify the signature before trying to parse it into actual data structures. 

This may be relevant for #367, but based on the shape of the JSON documents for V3 Webhooks, I'm not confident we will be able to implement a single function to deserialize the response and provide a single _useful_ data structure for the consumer to use.

V3 Webhooks do that thing where you need to have a two-phase deserialization, because the `event.data` field changes shape depending on the `event.event_type`. So I think they are ultimately going to have to do some of their own deserialization, most likely, with types we export. We could also offer functions for each type to make it easier in their code, since they'd just need to call `webhookv3.ExtractIncident(json.RawMessage)` or something.

That said, I do think it would make sense to offer an API to at least extract the top-level document to allow them the ability to do that second-phase deserialization, and maybe a helper one that combines them.

`VerifySignature()`
`Extract<Something>()`
`VerifyAndExtract<Something>()`

ChezCrawford:
> In general I try to qualify the validity of requests before I hand them to my actual handlers. So I'd probably verify the signature before trying to parse it into actual data structures. If memory were a constraint, I may combine them for efficiency purposes.

Ah I probably should have clarified.  When I proposed the two methods above, I was thinking that the second `func ExtractWebhookEvent(r *http.Request, secret string) (WebhookEvent, error)` would actually be more like a "VerifyAndExtract" method as you mentioned.

The `WebhookEvent` structure would likely contain a set of deserialized outer fields as well as the raw JSON message (that could then be used in that second stage of deserialization.

My thought was that second method would be the "all in one" approach (take the http request, verify the signature, return the partially deserialized event) and that the `VerifySignature` function would be more of a "utility" for folks who wanted more control over their HTTP request handling.  

To me, it seems like this current method is a strange mix of not quite utility yet not quite "all in one" either.  However, this may be something that is somewhat common in Go and just a little new to me.  If you're happy the way it is, i'll drop my approval.

PR #375, opened 2021-11-02
Make pd cli `go get`able

NautiluX:
This should enable users of the cli install/upgrade it with
```
go get -u github.com/PagerDuty/go-pagerduty/pd
```
I couldn't test it with my fork, though.


Makefile
```diff
@@ -10,7 +10,7 @@
 
 .PHONY: build
 build: build-deps
-	go build -mod=vendor -o pd ./command
+	go build -mod=vendor -o out/pd ./pd/main.go
```
theckman, line 13: Since we ignore `vendor/` it doesn't seem like there's value in having `-mod=vendor` on this command. Since you're already editing it, may as well remove that.

Makefile
```diff
@@ -20,7 +20,7 @@ build-deps:
 
 .PHONY: install
 install: build
-	cp pd $(GOROOT)/bin
+	go install -mod=vendor ./pd
```
theckman, line 23: Same comment about `-mod=vendor` here too.

README.md
```diff
@@ -5,15 +5,8 @@ go-pagerduty is a CLI and [go](https://golang.org/) client library for the [Page
...
-First, download the source code
 ```cli
-go get github.com/PagerDuty/go-pagerduty
-```
-
-Next build the application.
-```cli
-cd $GOPATH/src/github.com/PagerDuty/go-pagerduty
-make install
+go get -u github.com/PagerDuty/go-pagerduty/pd
```
theckman, line 9: This should be `github.com/PagerDuty/cmd/pd` per Go community consensus on project structure.

command/ability_check.go
```diff
@@ -1,4 +1,4 @@
-package main
+package cmd
```
theckman, line 1: Could this file and the rest remain as `package main` and instead be moved into `cmd/pd/`?

.gitignore
```diff
@@ -1,6 +1,5 @@
 bin/*
 *.swp
-pd
```
theckman, line 3: Could you add this back, and maybe also include the `out/*` directory that you now reference in the `Makefile`?

theckman (changes requested):
Thank you for this contribution. I think there are a few changes that would need to be made before this could be merged in.

Specifically, I'd like to see the size of the change reduced by just renaming the `command` directory to `cmd/pd` and update the docs / Makefile accordingly. This should accomplish making the binary `go get`able, while also largely keeping everything the same.

pd/main.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	. "github.com/PagerDuty/go-pagerduty/command"
```
theckman, line 7: I forgot to include this in the review comment. Please don't use dot imports, except for this one specific case:

https://github.com/golang/go/wiki/CodeReviewComments#import-dot

theckman:
@NautiluX wanted to check back in to see if you may have the time to make the requested changes.

theckman:
@NautiluX wanted to ping again to see if you may have the spare cycles to review the feedback.

Issue #339, opened 2021-06-30
APIError unmarshaling broken in 1.4


dsymonds:
When making some types of API calls, the error response fails to unmarshal into the JSON structures and so the error itself is lost to client code, who only gets a JSON error. In particular, I was creating an override in the past, and triggered this.

My apologies for not capturing the error itself (I worked around it, then lost my terminal history), but it should be easy to reproduce. Let me know if that's not the case and I'll try to go back into my local history.

dsymonds:
(Oh, to be clear, this was using v1.4.1, though I don't think any differences between v1.4.0 and v1.4.1 are relevant.

theckman:
@dsymonds I would need a reproduction case so that I can figure it out.

dsymonds:
I reproduced it with the `pd` command line tool in this repo (after fixing a wee bug in it).

Using an input file like:

```json
{
  "start": "2021-06-29T12:00:00Z",
  "end": "2021-06-29T18:00:00Z",
  "user": {...}
}
```

or with any other date in the past, you can get an API response like:

```json
{"message":"Invalid Override","code":2001,"errors":"Override must end in the future, Override must end after its start"}
```

and that doesn't unmarshal into the `APIErrorObject` struct, which has `Errors []string`.

theckman:
@dsymonds thank you for that. That `errors` key is supposed to be an array, at least according to their API documentation.

@stmcallister could you ping someone inside of PagerDuty to see whether that's expected? API documentation shows `errors` being an array but here it's a string.

https://developer.pagerduty.com/docs/rest-api-v2/errors/

stmcallister:
Interestingly, when I create an override with a raw API call, using dates in the past as you did, my errors object is an array of strings.

```
[
    {
        "status": 400,
        "errors": [
            "Override must end in the future",
            "Override must end after its start"
        ],
        "override": {
            "start": "2021-07-01T00:24:27Z",
            "end": "2021-06-02T04:00:00Z",
            "user": {
                "id": "PZACW3D",
                "type": "user_reference",
                "summary": "Bart J Simpson",
                "self": "https://api.pagerduty.com/users/PZACW3D",
                "html_url": "https://pdt-smcallister.pagerduty.com/users/PZACW3D"
            }
        }
    }
]
```

theckman:
I'm noticing that the output @stmcallister provided looks to be an array of objects, which may explain the failure. If so, I should be able to fix that.

[...]

PR #376, opened 2021-11-05
Add compatibility with API calls which return errors as a string

tstromberg:
With the intent of fixing #339

### DIFF
--- a/client.go
+++ b/client.go
@@ -83,6 +83,14 @@ type APIErrorObject struct {
 	Errors  []string `json:"errors,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


theckman:
@tstromberg thank you for taking the time to raise this Work In Progress PR. I took an initial look at it, but saw you had marked it as not being ready for review. This was something I did want to solve, in a slightly different way, so I raised #382.

We've gone with that approach (backported via #386) and have released `v1.4.3` which fixes #339. I'm going to close this PR at this time.

Thanks again!

tstromberg:
Excellent. Thank you!

PR #377, opened 2021-11-06
Add Service and User to LogEntry

theckman:
Now that the Service and User fields that are sent with Log Entries are
documented, we can add them to this library.

This also adds the `go-cmp` package to provide actionable test output when
comparing complex values.

Closes #201


stmcallister (approved):
Looks good! Thank you! 🎉

PR #380, opened 2021-11-09
Add missing required parameeter to ManageIncidentAlerts

theckman:
When making a call to the REST API[1] endpoint to manage alerts for incidents,
an HTTP header ("From") indicating who made the request is required. As a
result, these methods may not have ever worked properly.

So while we're breaking the API, let's remove the deprecated non-context version
to have fewer breaking changes in v2.

Closes #338

[1] https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE0NA-manage-alerts


stmcallister (approved):
Looks good! 🎉

PR #382, opened 2021-11-10
Handle unexpected type changes in PagerDuty REST API error responses

### LINKED ISSUE #339: APIError unmarshaling broken in 1.4

theckman:
The PagerDuty REST API documentation[1] indicates that the `errors` field of an
error response is an array of error strings. However, as shown in #339 the API
sometimes violates that specification and instead returns the `errors` field as
a string.

There is a PagerDuty customer support ticket open[2] to to address this issue, but
there is currently no ETA on the resolution. As such we are going to create this
workaround to properly parse the invalid responses and return the proper error
type to callers.

Closes #339

[1] https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTYz-errors
[2] https://tickets.pagerduty.com/hc/requests/341221

### DIFF
--- a/client.go
+++ b/client.go
@@ -83,6 +83,16 @@ type APIErrorObject struct {
 	Errors  []string `json:"errors,omitempty"`
 }
 
+// fallbackAPIErrorObject is a shim to solve this issue:
+// https://github.com/PagerDuty/go-pagerduty/issues/339
+//
+// TODO: remove when PagerDuty engineering confirms bugfix to the REST API
+type fallbackAPIErrorObject struct {
+	Code    int    `json:"code,omitempty"`
+	Message string `json:"message,omitempty"`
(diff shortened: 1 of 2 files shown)


theckman:
This should be backported to `release-1.4.x`, to create a v1.4.3.

stmcallister (approved):
This makes sense while the API shows the inconsistent behavior with returning errors. Thanks for adding this! 🎉

PR #383, opened 2021-11-12
Add omitempty JSON tag to User slice fields

### LINKED ISSUE #343: User slice fields missing omitempty
https://github.com/PagerDuty/go-pagerduty/blob/e23b94cea01f37da371fd9d7efb4e20ea0e36ff2/user.go#L34-L35
https://github.com/PagerDuty/go-pagerduty/blob/e23b94cea01f37da371fd9d7efb4e20ea0e36ff2/user.go#L37

These `User` fields should probably have their struct tags include `omitempty`, since you don't mutate them via the user object. There are dedicated API calls for managing contact methods, notification rules, and team member.

theckman:
These fields aren't actually updated by posting a User object (partial or full)
to the REST API, as other REST endpoints are used to mutate that data. As such,
there really isn't a reason to send these fields on any outbound requests that
make use of the `User` struct.

Closes #343

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -458,7 +458,7 @@ func TestClient_LastAPIRequest(t *testing.T) {
 	})
 
 	t.Run("integration", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


stmcallister (approved):
Makes sense. Thank you! 👍

PR #384, opened 2021-11-12
Add omitempty JSON tag to specific Service fields

### LINKED ISSUE #348: Should all fields in Service type be `omitempty
I am trying to just change the `alert_creation` type on a service (via `UpdateService`) and have noticed I cannot just set it as the JSON marshal wants to include `auto_resolve_timeout`, `acknowledgement_timeout`, `escalation_policy` (it isn't a pointer so `omitempty` doesn't work), `support_hours`, `scheduled_actions`. 

Is there a reason a few of these are lacking `omitempty`?

theckman:
Providing these fields on updates is optional, and so we should totally emit
them if they are empty versus accidentally unsetting them by sending a `null`
value.

Closes #348

### DIFF
--- a/service.go
+++ b/service.go
@@ -91,8 +91,8 @@ type Service struct {
 	APIObject
 	Name                    string                   `json:"name,omitempty"`
 	Description             string                   `json:"description,omitempty"`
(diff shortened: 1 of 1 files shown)


stmcallister (approved):
Good call! 👍

PR #385, opened 2021-11-13
Add support for adding email filters for Generic Email Integrations

### LINKED ISSUE #315: Add ability to create or update integrations with email filter rules
The pagerduty api supports fields for email integrations that allows specification from regex's and how to handle creation of incidents. Adding in structs for these properties can make it easier to manage these rules.

https://developer.pagerduty.com/api-reference/reference/REST/openapiv3.json/paths/~1services~1%7Bid%7D~1integrations~1%7Bintegration_id%7D/put

Example API call that can be used below.
```
{
  "integration": {
    "id": "ID",
    "type": "generic_email_inbound_integration",
    "summary": "Email",
    "self": "https://api.pagerduty.com/services/ID/integrations/ID",
    "html_url": "https://test.pagerduty.com/services/ID/integrations/ID",
    "name": "Email",
    "service": {
      "id": "ID",
      "type": "service_reference",
      "summary": "import-csv",
      "self": "https://api.pagerduty.com/services/ID",
      "html_url": "https://ukg-dev.pagerduty.com/service-directory/ID"
    },
    "created_at": "2021-03-18T13:52:41-04:00",
    "vendor": {
      "id": "PBPQJ1M",
      "type": "vendor_reference",
      "summary": "Email",
      "self": "https://api.pagerduty.com/vendors/ID",
      "html_url": null
    },
    "integration_email": "import-csv@test.com",
    "email_incident_creation": "on_new_email_subject",
    "email_filter_mode": "or-rules-email",
    "email_filters": [
      {
        "subject_mode": "always",
        "subject_regex": null,
        "body_mode": "match",
        "body_regex": "test",
        "from_email_mode": "always",
        "from_email_regex": null,
        "id": "ID"
      },
    ],
    "email_parsers": null,
    "email_parsing_fallback": null
  }
}
```

theckman:
This change adds support for adding different email filter rules to a generic
inbound email integration associated with a specific Service. This is done by
adding two new fields to the Integration struct.

While implementing this functionality I found that the API documentation from
PagerDuty lacked details about what the correct values are for specific fields,
and specifically what each value means. In fact, I actually had to go into the
PagerDuty UI to try and set up email filters, so that I could guess what the
different API values actually meant.

Because of that, I chose to create an enum-like type for each thing that wasn't
well documented to make it easier to consume via this client. That's resulted in
quite a lot of extra code, but the result is that anyone who wishes to consume
this area of the Go SDK should be able to figure it out how to use it without
much trouble.

Because of how the code grew, I also chose to move the Service
Integration-specific code into a new file (service_integration.go).

Closes #315


stmcallister (approved):
Breaking `service_integration` out to its own file makes sense. I've passed along your feedback on the API docs to our engineering team with the ask to let us know if your assumptions were incorrect. Thanks for adding this!

PR #391, opened 2021-11-14
Support adding conference bridge when creating or managing incidents

### LINKED ISSUE #373: Setting conference information in new incident
Hello, I'm looking to set the conference url and number when creating a new incident, however, I don't see a way to do that with this package? The [API supports](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE0MA-create-an-incident) it, but not the `CreateIncidentOptions` type.  And AFAICT the `Incident` type (which does have `ConferenceBridge`) is only used in responses from the API.

Am I overlooking something or would this be a feature request?

theckman:
Up until now you could view the conference bridge associated with an incident
when viewing its details via this package, but you could not add a conference
bridge to a new or existing incident. This change adds the ability to include a
conference bridge when creating a new incident, or when you're updating an
existing one.

Closes #373

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -114,8 +114,15 @@ type ListIncidentsOptions struct {
 
 // ConferenceBridge is a struct for the conference_bridge object on an incident
 type ConferenceBridge struct {
+	// ConferenceNumber is the phone number of the conference call for the
+	// conference bridge. Phone numbers should be formatted like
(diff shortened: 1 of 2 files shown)


stmcallister (approved):
👍 🎉 Thank you! Love the extensive commenting!

PR #392, opened 2021-11-14
Mark Type struct field deprecated, for incident creation + management

### LINKED ISSUE #390: When creating an incident, consumers shouldn't need to set the Type field
The `type` can only be incident, and so it should be set by default within the package and not exposed a consumer-settable option.

https://github.com/PagerDuty/go-pagerduty/blob/ad98290ed37bd905d8330a4e3c98d55b302d3f08/incident.go#L155

Since we can't remove it without it being a breaking change, we should deprecate it in preparation for v2.0.0.

theckman:
We should not expect consumers to set this value, as there is only a single
value that is valid (`incident`) based on how we use the API. As such, let's
deprecate the field and just set it in the methods that need it.

This also updates the tests to assert that we're automatically setting it
appropriately.

Closes #390

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -159,6 +159,12 @@ type createIncidentResponse struct {
 
 // CreateIncidentOptions is the structure used when POSTing to the CreateIncident API endpoint.
 type CreateIncidentOptions struct {
+	// Type is the type of API object this is.
+	//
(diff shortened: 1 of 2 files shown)


stmcallister (approved):
This makes sense. Again, thanks for the detailed comments! 👍

Issue #389, opened 2021-11-14
Validate that all structure formats and method signatures work with PagerDuty API


theckman:
We're making `v1.5.0` a special release, where we introduce breaking changes to fix issues within this package that probably never worked since being merged. Since we've fixed so many things so far, we should go over the entire package to make sure there are no outstanding issues.

Files to check:

- [x] ability.go
- [x] addon.go (#405)
- [x] analytics.go
- [x] business_service.go (#405)
- [x] change_events.go (#396)
- [x] escalation_policy.go (#396) (#405)
- [x] event.go
- [x] event_v2.go
- [x] extension.go (#414) (#405)
- [x] extension_schema.go (#405)
- [x] incident.go (#405) (#414) 
- [x] log_entry.go (#405) (#414)
- [x] maintenance_window.go (#405) (#414)
- [x] notification.go (#405) (#414)
- [x] on_call.go (#405)
- [x] priorites.go (#405)
- [x] ruleset.go (#414)
- [x] schedule.go (#405) (#414)
- [x] service.go (#405) (#414)
- [x] service_dependency.go
- [x] service_integration.go
- [x] tag.go (#405)
- [x] team.go (#405) (#414)
- [x] user.go (#405) (#414)
- [x] vendor.go (#405)
- [x] webhook.go

theckman:
Accomplished via #396, #405, and #414.

PR #396, opened 2021-11-16
Fix some mismatches between REST API and struct definitions

theckman:
This change includes the first few fixes while reviewing the repo for #389, and
does include one breaking change.

In `change_events.go`, there were a few optional fields that were not set to
`omitempty`, which means we were rendering them in the JSON even when not set.
This also reorders the `ChangeEventPayload` so that the `Summary` field is at
the top, since it's the only required field.

In `escalation_policy.go`, the `EscalationPolicy` type looked to have a field on
it that indicated if repeating the notification rules was enabled. However, this
field is not documented and when looking at the actual API response no such
field was present. Instead, the field to indicate the number of loops is set to
`0` when there is no repeating of the rules.

The removal of this field is the breaking change, although hopefully nobody was
using it. :)

Updates #389

### DIFF
--- a/change_events.go
+++ b/change_events.go
@@ -15,23 +15,23 @@ const changeEventPath = "/v2/change/enqueue"
 type ChangeEvent struct {
 	RoutingKey string             `json:"routing_key"`
 	Payload    ChangeEventPayload `json:"payload"`
-	Links      []ChangeEventLink  `json:"links"`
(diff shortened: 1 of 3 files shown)


stmcallister (approved):
If the `RepeatEnabled` field didn't do anything anyways I wouldn't think folks were using it. Thanks for cleaning up these fields! 👍

Issue #393, opened 2021-11-16
Reduce code duplication in analytics.go


theckman:
The three methods in `analytics.go` are like 99% the same, minus the last part of the API URL. This code is ripe for some DRYing by creating an unexported method, that accepts the final piece of as an argument and appends it to the URL.

https://github.com/PagerDuty/go-pagerduty/blob/2f646c3d8276a06bf62ab52f823a0ea2c0025ed2/analytics.go#L61-L115

theckman:
Fixed via #397.

PR #397, opened 2021-11-16
refactor: Reduce code duplication in analytics.go

t-junjie:
Refer to https://github.com/PagerDuty/go-pagerduty/issues/393

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 )
 
-const baseURL = "/analytics/metrics/incidents"
+const analyticsBaseURL = "/analytics/metrics/incidents"
 
 // AnalyticsRequest represents the request to be sent to PagerDuty when you want
 // aggregated analytics.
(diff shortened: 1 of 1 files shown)


analytics.go
```diff
@@ -2,8 +2,11 @@ package pagerduty
 
 import (
 	"context"
+	"fmt"
 )
 
+const baseURL = "/analytics/metrics/incidents"
```
theckman, line 8: Since `constants` are local to the package and not just this file, would you be able to make this constant name a little more precise? Maybe `analyticsBaseURL`?

analytics.go
```diff
@@ -60,48 +63,26 @@ type AnalyticsData struct {
...
+	return c.getAggregatedData(ctx, analytics, "teams")
+}
+
+func (c *Client) getAggregatedData(ctx context.Context, analytics AnalyticsRequest, endpoint string) (AnalyticsResponse, error) {
 	h := map[string]string{
 		"X-EARLY-ACCESS": "analytics-v2",
 	}
 
-	resp, err := c.post(ctx, "/analytics/metrics/incidents/teams", analytics, h)
+	URL := fmt.Sprintf("%s/%s", baseURL, endpoint)
```
theckman, line 84: This is an odd nit, I'll admit, but since capitalized identifiers are meant for things that are exported it feels a little unusual to have this be called `URL`. Could you rename it to `url` or even just `u`?

theckman (changes requested):
Thank you for this contribution! I did find two minor things that I requested changes on, but beyond that this should be good to go.

I had originally tagged the issue (#393) to be part of `v1.6.0`, but since you've opened this PR up so quickly I can make it part of the upcoming `v1.5.0` release once the changes are made.

t-junjie:
Thank you for reviewing my commits! 
I have made the appropriate changes.

theckman (approved):
Thank you again! I'll be merging this in shortly.

Issue #400, opened 2021-11-20
Add support for missing Incident fields


theckman:
* `assigned_via`
* `occurrence`

https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE0MQ-get-an-incident

PR #402, opened 2021-11-22
Missing incident fields

zonorti:
Not sure if a test makes sense here, just following the pattern.
Closes https://github.com/PagerDuty/go-pagerduty/issues/400

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -56,6 +56,15 @@ type Assignee struct {
 	Assignee APIObject `json:"assignee"`
 }
 
+// Occurrence
+type Occurrence struct {
+	Count     uint   `json:"count,omitempty"`
+	Frequency uint   `json:"frequency,omitempty"`
+	Category  string `json:"category,omitempty"`
+	Since     string `json:"since,omitempty"`
+	Until     string `json:"until,omitempty"`
+}
+
 // FirstTriggerLogEntry is the first LogEntry
 type FirstTriggerLogEntry struct {
 	CommonLogEntryField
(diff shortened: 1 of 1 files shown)


theckman (changes requested):
Thank you for this and your other PRs that you recently opened. I'll give them a proper once-over in the coming days, and them merge them once we're within the `v1.6.0` merge window. I don't have a precise ETA for that merge window yet, but my top priority is to try to get v1.5.0 out before the end of this year.

For this specific PR, would you be willing to squash the commits together? If they were two separate things being implemented I'd be okay with two commits, but since the second commit fixes an issue with the first I'd prefer they happen as a single commit.

zonorti:
@theckman Sure squashed it. Thanks for having beginner-friendly issues

incident.go
```diff
@@ -87,6 +96,8 @@ type Incident struct {
 	Body                 IncidentBody         `json:"body,omitempty"`
 	IsMergeable          bool                 `json:"is_mergeable,omitempty"`
 	ConferenceBridge     *ConferenceBridge    `json:"conference_bridge,omitempty"`
+	AssignedVia          string               `json:"assigned_via,omitempty"`
+	Occurrence           Occurrence           `json:"occurrence,omitempty"`
```
theckman, line 100: Does this always come back for an incident, or only some of the time. If it's only some of the time, might it make sense for this to be a pointer type (like `*ConferenceBridge`)?

zonorti: Looking at the API doc (which doesn't have this field in the example) it makes a lot of sense to me, thanks for pointing this out. 26b8fa2

theckman (approved):
Thank you for this PR as well. Similar to the other(s), it's currently slated for v1.6.0 but I may include it in v1.5.0.

Issue #318, opened 2021-03-23
Remove embedded APIObject from struct types, since it's not inheritance


theckman:
The V1 API currently tries to use embedding as inheritance by putting the `APIObject` inside of each type to avoid duplicating fields, and it's caused a few issues that will require us to make breaking changes to fix (#218 #316). We will need to find and fix these issues for v1.5.0, and for v2.0.0 we should remove the embeds completely and flatten the structures by hand.

whereswaldon:
So I'm just a random user with far less context than you have on this problem, so please take this as just an uninformed opinion: 
Your use of embedding here seems like a fine use case of composition, not inheritance. Grouping shared common fields into a subtype seems like a good idea.

From my outsider perspective, the problem is that the outer types that embed the `APIObject` duplicate some of its fields. I would think the fix is to remove the `Id`, `Type` and other shared fields from the outer types, not removing the `APIObject` type. I've read the issues that you linked, and I don't see a reason that removing the duplicated fields from the outer objects wouldn't resolve the issue.

Regardless, thank you for trying to maintain this codebase. I saw in another issue that you aren't even a PD employee anymore, and I would understand if you left it to rot. Maintainer's guilt is hard to shake, but sometimes you gotta to stay healthy.

So yeah, if you decide to try for a v2 API, I hope the above is helpful, and I'm sorry if I'm spewing nonsense because I don't understand the constraints you're solving for. Thanks again for your hard work!

theckman:
> Regardless, thank you for trying to maintain this codebase. I saw in another issue that you aren't even a PD employee anymore, and I would understand if you left it to rot. Maintainer's guilt is hard to shake, but sometimes you gotta to stay healthy.

@whereswaldon not even maintainer's guilt, my buddy started this project while he was there. I'm carrying his guilt, jk. 😂  

The half-truth is that I introduced Go at PagerDuty so I feel semi-responsible for him even starting the project (😛 ), but also I'm a user of it for my job at Netflix.

> Your use of embedding here seems like a fine use case of composition, not inheritance. Grouping shared common fields into a subtype seems like a good idea.

@whereswaldon thank you for sharing your thoughts on this, although I'm not sure I'm aligned with your perspective. Specifically, I've only ever seen composition used to describe a has-a relationship, while inheritance has been there to describe an is-a relationship.

[...]

PR #404, opened 2021-11-23
Add support for response_plays

petetanton:
Adding support for the response_plays API and enable adding response_plays to services

### DIFF
--- a/response_play.go
+++ b/response_play.go
@@ -1,0 +1,135 @@
+package pagerduty
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/google/go-querystring/query"
+)
+
+type ResponsePlay struct {
+	ID                 string          `json:"id,omitempty"`
+	Type               string          `json:"type,omitempty"`
+	Summary            string          `json:"summary,omitempty"`
+	Self               string          `json:"self,omitempty"`
+	HTMLURL            string          `json:"html_url,omitempty"`
+	Name               string          `json:"name,omitempty"`
+	Description        string          `json:"description"`
+	Team               *APIReference   `json:"team,omitempty"`
+	Subscribers        []*APIReference `json:"subscribers,omitempty"`
+	SubscribersMessage string          `json:"subscribers_message"`
+	Responders         []*APIReference `json:"responders,omitempty"`
+	RespondersMessage  string          `json:"responders_message"`
+	Runnability        *string         `json:"runnability"`
+	ConferenceNumber   *string         `json:"conference_number"`
+	ConferenceUrl      *string         `json:"conference_url"`
+	ConferenceType     *string         `json:"conference_type"`
+}
+
+type ListResponsePlaysResponse struct {
+	ResponsePlays []ResponsePlay `json:"response_plays"`
+}
+
+type ListResponsePlaysOptions struct {
+	FilterForManualRun bool   `url:"filter_for_manual_run,omitempty"`
+	Query              string `url:"query,omitempty"`
+}
+
+// ListResponsePlays lists existing response plays.
+func (c *Client) ListResponsePlays(ctx context.Context, o ListResponsePlaysOptions) (ListResponsePlaysResponse, error) {
+	var result ListResponsePlaysResponse
+
+	v, err := query.Values(o)
+	if err != nil {
+		return result, err
+	}
+
+	resp, err := c.get(ctx, "/response_plays?"+v.Encode())
+	if err != nil {
+		return result, err
+	}
+
+	if err = c.decodeJSON(resp, &result); err != nil {
+		return result, err
+	}
+
+	return result, nil
+}
+
+// CreateResponsePlay creates a new response play.
+func (c *Client) CreateResponsePlay(ctx context.Context, rp ResponsePlay) (ResponsePlay, error) {
+	d := map[string]ResponsePlay{
+		"response_play": rp,
+	}
+
+	resp, err := c.post(ctx, "/response_plays", d, nil)
+	return getResponsePlayFromResponse(c, resp, err)
+}
+
+// GetResponsePlay gets details about an existing response play.
+func (c *Client) GetResponsePlay(ctx context.Context, id string) (ResponsePlay, error) {
+	resp, err := c.get(ctx, "/response_plays/"+id)
+	return getResponsePlayFromResponse(c, resp, err)
+}
+
+// UpdateResponsePlay updates an existing response play.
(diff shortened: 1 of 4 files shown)


response_play.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// DeleteResponsePlayWithContext deletes an existing response play.
+func (c *Client) DeleteResponsePlayWithContext(ctx context.Context, id string) error {
+	_, err := c.delete(ctx, "/response_plays/"+id)
+	return err
+}
+
+// RunResponsePlayWithContext runs a response play on a given incident.
+func (c *Client) RunResponsePlayWithContext(ctx context.Context, rp ResponsePlay, incident Incident, from string) error {
```
theckman, line 69: I think this should take the response play ID as an argument, and not an entire `ResponsePlay`.

response_play.go
```diff
@@ -0,0 +1,111 @@
+package pagerduty
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+type ResponsePlay struct {
+	APIObject
```
theckman, line 10: Could you please remove this `APIObject` embed, and instead manually enumerate all the `APIObject` fields directly on this struct type? We're looking to remove this pattern in v2+, and so may as well try to avoid adding more cases of it. There's a little more context in #318 if you're curious.

response_play.go
```diff
@@ -0,0 +1,111 @@
...
+	Name        string        `json:"name,omitempty"`
+	Description string        `json:"description,omitempty"`
```
theckman, line 12: Because both of these are `string` and you have `omitempty`, a consumer of this package would never be able to provide an empty string to the API. This means you could never unset the name or description of a `ResponsePlay`.

According to the [API documentation](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE2Ng-create-a-response-play), neither the `name` or `description` are required. So is this the desired behavior?

theckman (changes requested):
Thank you so much for the contribution. I've requested a few changes on this PR. Some of things I requested apply to multiple areas of this PR, but I only commented once to avoid repetition, so please let me know if you have any questions or if any of the feedback is unclear.

petetanton:
@theckman thanks for your feedback!
I'll commit some changes and address your comments one-by-one

response_play.go
```diff
@@ -0,0 +1,111 @@
...
+	Team        *APIReference `json:"team,omitempty"`
+}
+
+type ListResponsePlaysResponse struct {
+	APIListObject
```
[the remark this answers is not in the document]
petetanton: Ah yes this is one of those strange PagerDuty API cases where the response looks paginated but is not

response_play.go
```diff
@@ -0,0 +1,111 @@
...
+	"fmt"
+	"net/http"
+)
+
+type ResponsePlay struct {
+	APIObject
+	Name        string        `json:"name,omitempty"`
+	Description string        `json:"description,omitempty"`
+	Team        *APIReference `json:"team,omitempty"`
+}
```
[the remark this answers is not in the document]
petetanton: Sorry I missed these - I will add them! :)

response_play.go
petetanton: I will remove `omitempty` from description but the `name` cannot be blank. The API Documentation is wrong, it is a required field

response_play.go
```diff
@@ -0,0 +1,135 @@
...
+	Description        string          `json:"description"`
+	Team               *APIReference   `json:"team,omitempty"`
+	Subscribers        []*APIReference `json:"subscribers,omitempty"`
+	SubscribersMessage string          `json:"subscribers_message"`
+	Responders         []*APIReference `json:"responders,omitempty"`
+	RespondersMessage  string          `json:"responders_message"`
+	Runnability        string          `json:"runnability"`
+	ConferenceNumber   string          `json:"conference_number"`
+	ConferenceUrl      string          `json:"conference_url"`
+	ConferenceType     string          `json:"conference_type"`
```
theckman, line 27: So I think this PR is basically ready to go, but I do have some reservations about this struct (not anything you did) and would love to hear your thoughts on them.

Specifically, because all of these fields are `string` and not `*string`, consumers can't do a partial update of the play without potentially erasing data. They will need to set all of the fields on the struct, or those fields will be zeroed out when the API call is made if they aren't required. 

A side effect of this is that consumers will almost always need to do a Get + Update operation, even if they are only updating a single field to a static value across all of their response plays. 

Should we make these a `*string` and not a `string`, and include `omitempty`, so that folks can do partial updates?

I'm not necessarily asking that we do this, but am instead asking your opinion on it and what you'd maybe do if you were deciding.

petetanton: Thanks @theckman  - yes I think that is a good suggestions, shall I make the changes and update the PR?

theckman: Sorry for the delay, was noodling on this.

Yeah, let's go with changing these up to be pointer values. :+1:

malderete:
@petetanton thanks for this contribution.
@theckman, could you review/merge this PR, please? I mean, is there anything else that need to be addressed in your opinion?

We're looking forward to use this feature :)

theckman:
I'll look at it shortly. Was focused on some of my own projects over the last few weeks.

theckman:
For some additional context, while I can merge the contributions of others into the `master` branch I cannot merge my own. So `v1.5.0`, and the release this was originally for (`v1.6.0`), are blocked on PagerDuty employees reviewing my outstanding PRs.

response_play.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// DeleteResponsePlay deletes an existing response play.
+func (c *Client) DeleteResponsePlay(ctx context.Context, id string) error {
+	_, err := c.delete(ctx, "/response_plays/"+id)
+	return err
+}
+
+// RunResponsePlay runs a response play on a given incident.
+func (c *Client) RunResponsePlay(ctx context.Context, rp string, incident Incident, from string) error {
```
theckman, line 93: Sorry for not explicitly calling this in an earlier review. I think it also makes sense to make this an `incidentID string` instead of `incident Incident`.

I think we could also make it a little more self-documenting, for those that use something like LSP to overlay method signatures as they are writing them. I'm thinking:

```Go
func (c *Client) RunResponsePlay(ctx context.Context, responsePlayID string, incidentID string, from string) error {
```

theckman (changes requested):

response_play.go
petetanton: no problem - I'll get this fixed up

petetanton: thanks for the clarification @theckman  - I have pushed a commit to change the signature :)

theckman (approved):

PR #405, opened 2021-11-25
Update pagination query to conform to API spec

theckman:
While looking for API shape mismatches for #389, I stumbled upon a major issue
with our pagination parameters across the entire package. This fixes that
discrepency, but is done via a pretty intrusive breaking change by removing the
embedded `APIListObject` from all `List*Options` types, replacing them with
three distinct fields: `Limit`, `Offset`, and `Total`.

For the traditional API pagination, there is a `total` URL query parameter that
tells the API to include the total count of records in the response. For the
query parameter it's a bool flag, with a value of `true` telling the API to
include the count.

When the response comes back from the API, there is a `total` field in the JSON
that is the total count of items in the collection. This value is only set if
that `total` query parameter is set to `true` because it's an expensive
operation that PagerDuty recommends you don't enable.

The problem is we tried to reuse the `APIListObject` struct type for both the
pagination URL query parameters, and the pagination fields in the JSON response
body. The issue with doing that is for requests the `Total` field needs to be a
bool, but within the `APIListObject` it's a `uint` because it was really only designed
to be used for response body parsing.

So since we're intending to remove the embedded struct types in v2 (#318), it
feels like the best course of action here is to move forward with that plan for
these paginated query parameter struct types so that we can conform to the
PagerDuty API spec.

Related to #318
Related to #389

### DIFF
--- a/priorites.go
+++ b/priorites.go
@@ -2,6 +2,8 @@ package pagerduty
 
 import (
 	"context"
+
+	"github.com/google/go-querystring/query"
 )
 
 // PriorityProperty is a single priorty object returned from the Priorities endpoint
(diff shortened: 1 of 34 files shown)


priorites.go
```diff
@@ -11,28 +13,56 @@ type PriorityProperty struct {
 	Description string `json:"description"`
 }
 
-// Priorities repreents the API response from PagerDuty when listing the
-// configured priorities.
-type Priorities struct {
+// ListPrioritiesResponse repreents the API response from PagerDuty when listing
+// the configured priorities.
+type ListPrioritiesResponse struct {
```
theckman, line 18: I could be convinced to not make this change, but it feels like a pretty low cost to help make this area of the code more consistent when we're already introducing breaking changes.

schedule.go
```diff
@@ -295,7 +309,6 @@ func (c *Client) DeleteOverrideWithContext(ctx context.Context, scheduleID, over
 
 // ListOnCallUsersOptions is the data structure used when calling the ListOnCallUsers API endpoint.
 type ListOnCallUsersOptions struct {
-	APIListObject
```
theckman, line 298: For all of these `APIListObject` deletes in this file, it looks like they were added to these types by mistake. They are not part of the API specification.

vendor.go
```diff
@@ -33,8 +33,24 @@ type ListVendorResponse struct {
 
 // ListVendorOptions is the data structure used when calling the ListVendors API endpoint.
 type ListVendorOptions struct {
-	APIListObject
-	Query string `url:"query,omitempty"`
```
theckman, line 37: `query` does not appear in the API documentation for this call. So I removed it.

priorites.go
stmcallister: Agreed. I'm a big fan of consistency.

stmcallister (approved):
This makes sense. Thanks for making these changes! 👍

priorites.go
theckman: I've opened #416 as I remember a way to do this in a non-breaking way. :+1:

Issue #270, opened 2021-02-12
Investigating returning non-pointer types from methods / functions


theckman:
There are some pointer values returned from methods, like `ListIncidents`, where it's not clear why the value is returned as a pointer value. We should look at changing these return types in v2.0.0 to be their non-pointer types.

PR #406, opened 2021-12-06
add createStatusUpdate

kkawamura:
API has the action to create status updates. But `go-pagerduty` does not have the function for this yet.
I add some new functions to solve this issue.
https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE1Mw-create-a-status-update-on-an-incident

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -653,4 +653,35 @@ func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID
 	return &result, nil
 }
 
-/* TODO: Create Status Updates */
+// IncidentStatusUpdate is a status update for the specified incident.
+type IncidentStatusUpdate struct {
+	ID        string	`json:"id"`
+	Message   string	`json:"message"`
+	CreatedAt string	`json:"created_at"`
+	Sender    APIObject	`json:"sender"`
+}
+
+// CreateIncidentStatusUpdate creates a new status update for the specified incident.
+func (c *Client) CreateIncidentStatusUpdate(ctx context.Context, id string, from string, message string) (IncidentStatusUpdate, error) {
+	d := map[string]string{
+		"message": message,
+	}
+
+	h := map[string]string{
+		"From": from,
+	}
+
+	resp, err := c.post(ctx, "/incidents/"+id+"/status_updates", d, h)
+	if err != nil {
+		return IncidentStatusUpdate{}, err
+	}
+
+	var result struct {
+		IncidentStatusUpdate IncidentStatusUpdate `json:"status_update"`
+	}
+	if err = c.decodeJSON(resp, &result); err != nil {
+		return IncidentStatusUpdate{}, err
+	}
+
+	return result.IncidentStatusUpdate, nil
+}
--- a/incident_test.go
+++ b/incident_test.go
@@ -1,6 +1,7 @@
 package pagerduty
 
 import (
+	"context"
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
@@ -739,6 +740,58 @@ func TestIncident_ManageAlerts(t *testing.T) {
 					ID: "1",
 				},
 			},
+		},
+	}
+
+	if err != nil {
+		t.Fatal(err)
+	}
+	testEqual(t, want, res)
+}
+
+// CreateIncidentStatusUpdate
+func TestIncident_CreateIncidentStatusUpdate(t *testing.T) {
+	setup()
+	defer teardown()
+
+	wantId := "1"
+	wantFrom := "foo@bar.com"
+	wantMessage := "foo"
+
+	mux.HandleFunc("/incidents/1/status_updates", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+
+		if gotFrom := r.Header.Get("From"); gotFrom != wantFrom {
+			t.Errorf("From HTTP header = %q, want %q", gotFrom, wantFrom)
+		}
+
+		body, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		var data map[string]string
+		if err := json.Unmarshal(body, &data); err != nil {
+			t.Fatal(err)
+		}
+		o, ok := data["message"]
+		if !ok {
+			t.Fatal("map does not have message key")
+		}
+		if o != wantMessage {
+			t.Errorf("message = %q, want %q", o, wantMessage)
+		}
+
+		_, _ = w.Write([]byte(`{"status_update": {"id": "1", "message": "foo", "sender": {"summary": "foo@bar.com", "type": "user_reference"}}}`))
+	})
+	client := defaultTestClient(server.URL, "foo")
+	res, err := client.CreateIncidentStatusUpdate(context.Background(), wantId, wantFrom, wantMessage)
+	want := IncidentStatusUpdate{
+		ID:      wantId,
+		Message: wantMessage,
+		Sender: APIObject{
+			Summary: wantFrom,
+			Type: "user_reference",
 		},
 	}
 


incident.go
```diff
@@ -653,4 +653,57 @@ func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID
...
+	Sender    APIObject `json:"sender,omitempty"`
+}
+
+// CreateIncidentStatusUpdateResponse is returned from the API as a response to creating an incident status update.
+type CreateIncidentStatusUpdateResponse struct {
+	IncidentStatusUpdate IncidentStatusUpdate `json:"status_update"`
+}
+
+// CreateIncidentStatusUpdateWithContext creates a new status update for the specified incident.
+func (c *Client) CreateIncidentStatusUpdateWithContext(ctx context.Context, id string, statusUpdate IncidentStatusUpdate) (*IncidentStatusUpdate, error) {
```
theckman, line 670: I think there are a few things we should change with this method signature before merging.

* It should be renamed to `CreateIncidentStatusUpdate`.
  * The methods that are named `...WithContext` were added so we could introduce `context.Context` support without breaking API compatibility. New methods should accept the `context.Context` and not use this naming scheme.
* Instead of accepting a `statusUpdate IncidentStatusUpdate`, I'd rather see this take a `from, message string`.
  * This will make it work more like other methods that require a `from` parameter, and it more closely matches the [REST API](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE1Mw-create-a-status-update-on-an-incident) which only needs the string value of the `message`, and not the full `IncidentStatusUpdate` JSON object.
* This should not return a pointer (`*IncidentStatusUpdate`) and instead return a concrete `IncidentStatusUpdate` value.

incident.go
```diff
@@ -653,4 +653,57 @@ func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID
...
+		return nil, err
+	}
+
+	return &result.IncidentStatusUpdate, nil
+}
+
+// CreateIncidentStatusUpdateWithResponse creates a new status update for the specified incident.
+//
+// Deprecated: Use CreateIncidentStatusUpdateWithContext instead.
+func (c *Client) CreateIncidentStatusUpdateWithResponse(id string, statusUpdate IncidentStatusUpdate) (*IncidentStatusUpdate, error) {
```
theckman, line 695: What's the purpose of this method?

incident.go
```diff
@@ -653,4 +653,57 @@ func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID
...
+	return c.CreateIncidentStatusUpdateWithContext(context.Background(), id, statusUpdate)
+}
+
+// CreateIncidentStatusUpdate creates a new status update for the specified incident.
+//
+// Deprecated: Use CreateIncidentStatusUpdateWithContext instead.
+func (c *Client) CreateIncidentStatusUpdate(id string, statusUpdate IncidentStatusUpdate) error {
```
theckman, line 702: I believe this method should be removed.

incident.go
```diff
@@ -653,4 +653,57 @@ func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID
...
+// IncidentStatusUpdate is a status update for the specified incident.
+type IncidentStatusUpdate struct {
+	ID        string    `json:"id,omitempty"`
+	Message   string    `json:"message,omitempty"`
+	CreatedAt string    `json:"created_at,omitempty"`
+	Sender    APIObject `json:"sender,omitempty"`
+}
+
+// CreateIncidentStatusUpdateResponse is returned from the API as a response to creating an incident status update.
+type CreateIncidentStatusUpdateResponse struct {
```
theckman, line 665: Since we only use this to deserialize the JSON response, and don't actually return it to consumers (or accept it in any methods), does this need to be exported?

incident_test.go
```diff
@@ -747,3 +747,51 @@ func TestIncident_ManageAlerts(t *testing.T) {
 	}
 	testEqual(t, want, res)
 }
+
+// CreateIncidentStatusUpdate
+func TestIncident_CreateIncidentStatusUpdate(t *testing.T) {
```
theckman, line 752: This test does not appear to assert that the request was formatted as expected before sending it to the API (HTTP endpoint). Could you expand this test to actually assert that the request looks correct for both the `From` header and the `message` in the request body?

Maybe this other test can be of inspiration, with some changes being needed to work for this specific request: https://github.com/PagerDuty/go-pagerduty/blob/9190a3b061127b6b934a72c442843a3ee51bec61/incident_test.go#L54-L81

theckman (changes requested):
Thank you for this contribution! There are a few changes that will need to be made to this PR before it's ready to be merged in.

Also, could you please squash your commits down to a single commit while making these changes?

incident.go
kkawamura: I was trying to prepare the method sets like other old methods. But I can understand that we don't need this because of your helpful comments. I'll remove this.

incident.go
kkawamura: Okay. I'll remove this.

incident.go
```diff
@@ -653,4 +653,57 @@ func (c *Client) ManageIncidentAlertsWithContext(ctx context.Context, incidentID
 	return &result, nil
 }
 
-/* TODO: Create Status Updates */
+// IncidentStatusUpdate is a status update for the specified incident.
+type IncidentStatusUpdate struct {
+	ID        string    `json:"id,omitempty"`
+	Message   string    `json:"message,omitempty"`
+	CreatedAt string    `json:"created_at,omitempty"`
+	Sender    APIObject `json:"sender,omitempty"`
```
theckman, line 661: I apparently forgot to add this comment to my review before submitting it...

Since we don't use this type to serialize JSON, and only to deserialize it, I don't think the `omitempty` tags are needed on this type.

incident.go
kkawamura: Ah. OK. You mean, 
```
CreateIncidentStatusUpdate(ctx context.Context, id string, from string, message string)
```

This is the best. Right?

theckman: Yeah, that looks good to me. The complete header would end up being:

```Go
func (c *Client) CreateIncidentStatusUpdateWithContext(ctx context.Context, id string, from string, message string) (*IncidentStatusUpdate, error)
```

theckman: Yeah, that looks good to me. The complete header would end up being:

```Go
func (c *Client) CreateIncidentStatusUpdate(ctx context.Context, id string, from string, message string) (*IncidentStatusUpdate, error)
```

theckman: Yeah, that looks good to me. The complete header would end up being:

```Go
func (c *Client) CreateIncidentStatusUpdate(ctx context.Context, id string, from string, message string) (IncidentStatusUpdate, error)
```

kkawamura: >This should not return a pointer (*IncidentStatusUpdate) and instead return a concrete IncidentStatusUpdate value.

Sorry, why do you think that only this thread should not return a pointer?

theckman: The primary purpose of pointers are to permit shared mutability of some value, with a secondary purpose to be to differentiate between there being no value and a zero value (think `*string` when using `omitempty` with JSON).

In this case, were are returning a result from the API. The client isn't going to update the value at all once it's returned, and so there isn't a need for shared mutability. The value probably won't be mutated at all, and just read.

To my second purpose, we don't need to differentiate between no value and the zero value, because us returning an `error` handles that. If the `error` is not nil, the other returned values should be discarded.

kkawamura: Thank you for your reply. I agree with your point.

But I mean, other similar method (e.g. CreateIncidentNotes etc) return a pointer. If only this method return value, it might be confused for developers, I thought. 

Ok. let me change to use value instead of a pointer.

incident.go
kkawamura: I removed omitempty tag. Thanks.

incident_test.go
kkawamura: Thanks. Done.

kkawamura:
@theckman Thank you for your review. Please check the new code if you have time.

incident.go
theckman: Yeah. That is definitely true and I appreciate you calling that out.

For v2+ I plan to review all of those return types, and make sure we only return a pointer when it makes sense. I created an issue a few months back to track this: https://github.com/PagerDuty/go-pagerduty/issues/270.

So until v2 comes along, I'm leaning in to some inconsistency to reduce the number of breaking changes we will need to make in the future.

Thank you for changing this!

theckman (approved):
Thank you!

Issue #394, opened 2021-11-16
Add support for fetching Audit Records


theckman:
https://developer.pagerduty.com/api-reference/b3A6Mjc0ODExNA-list-audit-records

theckman:
@t-junjie that would be great, I appreciate it! I do have a few suggestions which may be helpful for when you do get to implementing it.

The first is that for new methods we're adding to the project, we're not creating new `WithContext()` variations. Instead, the new base method is always accepting a `ctx context.Context` as its first argument. So I think the method definition would be something like `ListAuditRecords(ctx context.Context, o ListAuditRecordOptions) (ListAuditRecordsResponse, error)`.

The second is that for these types of paginated API calls, for the future we're looking to have two separate methods. The first, `ListAuditRecords`, which makes a single API call and returns the response back (including pagination inputs for the next call). The other is one which handles pagination, and just returns the full set of entities back to the caller. So that would be `ListAuditRecordsPaginated`, and I talk about that change in this issue: https://github.com/PagerDuty/go-pagerduty/issues/295.

I also think we should be moving to a mode where we allow consumers to provide an `include` function in the `Paginated` variants, so that they can filter out any records they don't want before the full slice is returned to them. I discuss that in this issue: https://github.com/PagerDuty/go-pagerduty/issues/296.

So the result is I think we'd end up with two new methods:

```Go
func (c *Client) ListAuditRecords(ctx context.Context, o ListAuditRecordOptions) (ListAuditRecordsResponse, error) {}

func (c *Client) ListAuditRecordsPaginated(ctx context.Context, o ListAuditRecordOptions, include func(AuditRecord) bool) ([]AuditRecord, error) {}
```

I thought this information might be more helpful to share now instead of during a PR review. Please let me know if you have any questions.

theckman:
I forgot to include this above, here is an example of another `Paginated` method in the codebase. One note, it was built before I'd considered adding the `include` function (#296):

https://github.com/PagerDuty/go-pagerduty/blob/2f646c3d8276a06bf62ab52f823a0ea2c0025ed2/service.go#L156-L185

t-junjie:
Thanks for the suggestions and example! 
I might be a little slow, but I will get started on this and follow up with you if/when I have any questions.

t-junjie:
I took a look at the example in #296, as well as the above example. 
I am confused as to whether I should be using `pagedGet` or  `get` to retrieve the audit records. 

[...]

PR #408, opened 2021-12-22
feat: Add support for fetching Audit Records

t-junjie:
See issue [394](https://github.com/PagerDuty/go-pagerduty/issues/394)

### DIFF
--- a/audit.go
+++ b/audit.go
@@ -1,0 +1,167 @@
+package pagerduty
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/google/go-querystring/query"
+)
+
+const auditBaseURL = "/audit/records"
+
+// ListAuditRecordsOptions is the data structure used when calling the
+// ListAuditRecords API endpoint.
+type ListAuditRecordsOptions struct {
+	Actions              []string `url:"actions,omitempty,brackets"`
+	ActorID              string   `url:"actor_id,omitempty"`
+	ActorType            string   `url:"actor_type,omitempty"`
+	Cursor               string   `url:"cursor,omitempty"`
+	Limit                uint     `url:"limit,omitempty"`
+	MethodTruncatedToken string   `url:"method_truncated_token,omitempty"`
+	MethodType           string   `url:"method_type,omitempty"`
+	RootResourcesTypes   []string `url:"root_resources_types,omitempty,brackets"`
+	Since                string   `url:"since,omitempty"`
+	Until                string   `url:"until,omitempty"`
+}
+
+// ListAuditRecordsResponse is the response data received when calling the
+// ListAuditRecords API endpoint.
+type ListAuditRecordsResponse struct {
+	Records []AuditRecord `json:"records,omitempty"`
+	// ResponseMetaData is not a required field in the pagerduty API response,
+	// using a pointer allows us to not marshall an empty ResponseMetaData struct
+	// into a JSON.
+	ResponseMetaData *ResponseMetaData `json:"response_metadata,omitempty"`
+	Limit            uint              `json:"limit,omitempty"`
+	// NextCursor is an  opaque string that will deliver the next set of results
+	// when provided as the cursor parameter in a subsequent request.
+	// A null value for this field indicates that there are no additional results.
+	// We use a pointer here to marshall the string value into null
+	// when NextCursor is an empty string.
+	NextCursor *string `json:"next_cursor"`
+}
+
+// AuditRecord is a audit trail record that matches the query criteria.
+type AuditRecord struct {
+	ID               string           `json:"id,omitempty"`
+	Self             string           `json:"self,omitempty"`
+	ExecutionTime    string           `json:"execution_time,omitempty"`
+	ExecutionContext ExecutionContext `json:"execution_context,omitempty"`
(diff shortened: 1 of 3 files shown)


audit.go
```diff
@@ -0,0 +1,162 @@
...
+	NextCursor       string           `json:"next_cursor,omitempty"`
+}
+
+// AuditRecord is a audit trail record that matches the query criteria.
+type AuditRecord struct {
+	ID               string           `json:"id,omitempty"`
+	Self             string           `json:"self,omitempty"`
+	ExecutionTime    string           `json:"execution_time,omitempty"`
+	ExecutionContext ExecutionContext `json:"execution_context,omitempty"`
+	Actors           []Actor          `json:"actors,omitempty"`
```
theckman, line 43: Let's switch this to be `[]APIObject`, I'll explain why in my comment on the `Actor` type definition...

audit.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// ExecutionContext contains information about the action execution context.
+type ExecutionContext struct {
+	RequestID     string `json:"request_id,omitempty"`
+	RemoteAddress string `json:"remote_address,omitempty"`
+}
+
+// Actor is the actor who carried out the action.
+type Actor APIObject
```
theckman, line 62: I think we should avoid this becoming a named type. If PagerDuty ever adds any functionality in the future, where something else is referred to as an "actor", we can no longer use this name without a breaking change. I think using it directly as an `APIObject` is more clear to consumers that it's not a hydrated object.

audit.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Resource contains information on the resource that was audited.
+type Resource APIObject
```
theckman, line 65: I have the same thoughts here as I do about `Actor`.

audit.go
```diff
@@ -0,0 +1,162 @@
...
+type Reference struct {
+	Name        string      `json:"name,omitempty"`
+	Description string      `json:"description,omitempty"`
+	Added       []APIObject `json:"added,omitempty"`
+	Removed     []APIObject `json:"removed,omitempty"`
+}
+
+// ListAuditRecords lists audit trial records matching provided query params
+// or default criteria.
+func (c *Client) ListAuditRecords(ctx context.Context, o ListAuditRecordsOptions) (*ListAuditRecordsResponse, error) {
```
theckman, line 103: I think per issue #270 we should return `(ListAuditRecordsResponse, error)`.

audit.go
```diff
@@ -0,0 +1,162 @@
...
+	if err = c.decodeJSON(resp, &result); err != nil {
+		return nil, err
+	}
+
+	return &result, nil
+
+}
+
+// ListAuditRecordsPaginated lists audit trial records matching provided query
+// params or default criteria, processing paginated responses.
```
theckman, line 125: Could you add these to sentences to the end of this comment:

```
The include function decides whether or not to include a specific AuditRecord in the final result. If the include function is nil, all audit records from the API are included by default.
```

The rest looks good!

client.go
```diff
@@ -656,3 +656,53 @@ func (c *Client) pagedGet(ctx context.Context, basePath string, handler response
...
+		if len(c.NextCursor) == 0 {
+			break
+		}
+
+		// Otherwise, update next to parse more results.
+		next = c.NextCursor
+	}
+
+	return nil
+}
```
theckman, line 708: Thank you for doing all of this! 👍

theckman (changes requested):
Thank you so much! This looks almost ready, with a few small requested changes.

Could you also squash your commits down to a single one just to help keep the history cleaner once it's merged?

t-junjie:
I pulled in changes from `master` and made the relevant changes as requested above. 
It looks like I overwrote some of the commits when I squashed them when I did a `git rebase`, causing the previous review to be "outdated". Sorry about that

theckman:
The tests are failing:

```
=== RUN   TestAudit_List
    audit_test.go:94: values not equal (-want / +got):
          interface{}(
        - 	&pagerduty.ListAuditRecordsResponse{
        - 		Records: []pagerduty.AuditRecord{
        - 			{
        - 				ID:               "PDRECORDID4_UPDATED_USERS_NOTIFI"...,
        - 				ExecutionTime:    "2020-06-04T15:30:16.272Z",
        - 				ExecutionContext: pagerduty.ExecutionContext{...},
        - 				Actors:           []pagerduty.APIObject{...},
        - 				...
        - 			},
        - 		},
        - 		Limit: 10,
        - 	},
        + 	pagerduty.ListAuditRecordsResponse{
        + 		Records: []pagerduty.AuditRecord{
        + 			{
        + 				ID:               "PDRECORDID4_UPDATED_USERS_NOTIFI"...,
        + 				ExecutionTime:    "2020-06-04T15:30:16.272Z",
        + 				ExecutionContext: pagerduty.ExecutionContext{...},
        + 				Actors:           []pagerduty.APIObject{...},
        + 				...
        + 			},
        + 		},
        + 		Limit: 10,
        + 	},
          )
--- FAIL: TestAudit_List (0.00s)
````

t-junjie:
I made the appropriate changes and the test has passed.

theckman (approved):

PR #414, opened 2022-01-16
Second batch of fixes for API incompatibilities

theckman:
Updates #389


stmcallister (approved):
Another good looking PR! Especially appreciate the commenting that you've added as well! 🎉

PR #416, opened 2022-01-18
Find a way to gracefully avoid one breaking change in #405

theckman:
In #405 I opted to rename one of the types to be more consistent, and in the
moment totally forgot about type aliases. A type alias allows you to rename a
type in a non-breaking way, by converting the old name to be an alias to the new
one.

This change does that, and effectively removes one breaking change from v1.5.0.

I also do the sane thing to unify the Priority type into one.

### DIFF
--- a/priorites.go
+++ b/priorites.go
@@ -6,19 +6,24 @@ import (
 	"github.com/google/go-querystring/query"
 )
 
-// PriorityProperty is a single priorty object returned from the Priorities endpoint
-type PriorityProperty struct {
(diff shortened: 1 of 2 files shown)


theckman:
https://github.com/PagerDuty/go-pagerduty/pull/405#discussion_r756650209

stmcallister (approved):
Nice! TIL about aliasing! I like this solution better! Thanks! 👍 🌮 🎉

PR #417, opened 2022-01-19
Fix linter issues introduced by final PR merges

theckman:


theckman:
These have not been released yet, so it's safe to rename.

stmcallister (approved):
Cool. 👍

PR #419, opened 2022-01-24
Return custom error type for Events API (V2) calls

theckman:
This was originally opened via #368, but the original author didn't seem able to be able address the last round of feedback. So I squashed all of their commits into one, and then added a second commit to address my feedback from that PR. 

Closes #368


stmcallister (approved):
Thanks for finishing this one up! 👍

PR #423, opened 2022-02-15
Add ResolvedAt for IncidentAlert

gringus:
Add missing property.


theckman:
@gringus PagerDuty does not document this field, and so we can't accept this right now. Would you be willing to shoot them a support request to update the docs? (I'm asking because I don't work there, and have no different way to escalate).

gringus:
Will check with them. Thanks.

PR #425, opened 2022-02-16
Update client.go

priyankasekhar:
Adding the 'deleted_at' variable for a user. This will return the date/time a user was deleted from an escalation policy and is returned by /oncalls Pagerduty API today.

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,6 +51,7 @@ type APIObject struct {
 	Summary string `json:"summary,omitempty"`
 	Self    string `json:"self,omitempty"`
 	HTMLURL string `json:"html_url,omitempty"`
+	DeletedAt string `json:"deleted_at,omitempty"`
 }
 
 // APIListObject are the fields used to control pagination when listing objects.


client.go
```diff
@@ -51,6 +51,7 @@ type APIObject struct {
 	Summary string `json:"summary,omitempty"`
 	Self    string `json:"self,omitempty"`
 	HTMLURL string `json:"html_url,omitempty"`
+	DeletedAt string `json:"deleted_at,omitempty"`
```
theckman, line 54: @priyankasekhar as the majority of API endpoints do not return this field, I don't think it's appropriate to place it on that type.

theckman (changes requested):
Could you point me to where PagerDuty documents the `deleted_at` field for the API call you're talking about? I checked here, but did not see it: https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE2Mw-list-all-of-the-on-calls

Either way, I don't believe where we are trying to add this field is the right place as not all objects include `deleted_at`.

client.go
priyankasekhar: @theckman You're right, ideally when all the users part of an escalation policy are active in Pagerduty, the deleted_at variable is not present in the response. It's only returned when a user is marked 'XYZ -inactive'. Their documentation doesn't really reference to this variable. But here's an example of such a scenario:
Screen Shot 2022-02-16 at 3 44 50 PM

priyankasekhar: Maybe it'll make more sense to add this variable here: https://github.com/PagerDuty/go-pagerduty/blob/master/user.go#L38 ?

priyankasekhar: This is the documentation they shared with me on "what happens when a user is deleted and that user is currently on an escalation policy". Although It doesn't mention the deleted_at variable : [Gaps in Escalation Policies and Scheduling Settings](https://support.pagerduty.com/docs/offboarding#gaps-in-escalation-policies-and-schedules-settings).
Does this help at all?

theckman: I don't know that it makes sense to add it to the User object, as I'm not sure I've seen `deleted_at` get mentioned for users anywhere else. It'd be helpful to know where-all PagerDuty can return the `deleted_at` field.

That being visible in the JSON output doesn't help, unfortunately. Per policy set on this project by PagerDuty employees, the field needs to be listed in the documentation for it to be supported by this package. 

I don't personally have an inside way to ask for it to be fixed, and generally requires someone reach out to `support@pagerduty.com` to ask they do it.

PR #429, opened 2022-03-06
Add mechanism to list more than 100 incidents

xccatana:
By default we can reach only 100 elements from a given range, if we have more than 100 elements
between `Since` and `Until`, we need to use an offeset to get the 101st element and so on.

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -3,6 +3,7 @@ package pagerduty
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/google/go-querystring/query"
 )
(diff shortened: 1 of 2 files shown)


xccatana:
I will implement it for v1.4.2. If you think is good for the project, I will reopen it.

incident.go
```diff
@@ -183,6 +188,37 @@ func (c *Client) ListIncidentsWithContext(ctx context.Context, o ListIncidentsOp
 	return &result, nil
 }
 
+// ListIncidentsPaginated lists existing services processing paginated responses
+func (c *Client) ListIncidentsPaginatedWithContext(ctx context.Context, o ListIncidentsOptions) ([]Incident, error) {
```
theckman, line 192: I don't think this method signature aligns with future plans of the project. The `WithContext` should be removed from the name, as this is a new method and we have no reason to support a non-context APIs for new things. Also, it should include accepting a function that allows filtering of unwanted items.

See https://github.com/PagerDuty/go-pagerduty/issues/295 and https://github.com/PagerDuty/go-pagerduty/issues/296 for more info.

incident.go
```diff
@@ -163,6 +164,10 @@ func (c *Client) ListIncidents(o ListIncidentsOptions) (*ListIncidentsResponse,
 	return c.ListIncidentsWithContext(context.Background(), o)
 }
 
+func (c *Client) ListIncidentsPaginated(o ListIncidentsOptions) ([]Incident, error) {
```
theckman, line 167: This method should be removed.

theckman (changes requested):
Because this is a new feature it would not be released as part of `v1.4.x` or `v1.5.x`, and could be included in `v1.6.0`. I think there's value in adding it, but I do have some minor requested changes before it could be merged. 👍

PR #432, opened 2022-03-16
Add slack connection apis

GiedriusS:
Add code for working with the relatively new Slack connection APIs as defined here: https://developer.pagerduty.com/api-reference/b3A6MTEyMDk1NDU-list-slack-connections and in other pages.

### DIFF
--- a/slack_connection.go
+++ b/slack_connection.go
@@ -1,0 +1,128 @@
+package pagerduty
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/google/go-querystring/query"
+)
+
+// SlackConnectionConfig is the configuration of a Slack connection as per the documentation
+// https://developer.pagerduty.com/api-reference/c2NoOjExMjA5MzMy-slack-connection.
+type SlackConnectionConfig struct {
+	Events     []string `json:"events"`
+	Urgency    *string  `json:"urgency"`
+	Priorities []string `json:"priorities"`
+}
+
+// SlackConnection is an entity that represents a Slack connections as per the
+// documentation https://developer.pagerduty.com/api-reference/c2NoOjExMjA5MzMy-slack-connection.
+type SlackConnection struct {
+	SourceID   string `json:"source_id"`
+	SourceName string `json:"source_name"`
+	SourceType string `json:"source_type"`
+
+	ChannelID        string `json:"channel_id"`
+	ChannelName      string `json:"channel_name"`
+	NotificationType string `json:"notification_type"`
+
+	Config SlackConnectionConfig `json:"config"`
+}
+
+// SlackConnectionObject is an API object returned by getter functions.
+type SlackConnectionObject struct {
+	SlackConnection
+	APIObject
+}
+
+// ListSlackConnectionsResponse is an API object returned by the list function.
+type ListSlackConnectionsResponse struct {
+	Connections []SlackConnection `json:"slack_connections"`
+	APIListObject
+}
+
+// ListSlackConnectionsOptions is the data structure used when calling the ListSlackConnections API endpoint.
+type ListSlackConnectionsOptions struct {
+	// Limit is the pagination parameter that limits the number of results per
+	// page. PagerDuty defaults this value to 50 if omitted, and sets an upper
+	// bound of 100.
+	Limit uint `url:"limit,omitempty"`
+
+	// Offset is the pagination parameter that specifies the offset at which to
+	// start pagination results. When trying to request the next page of
+	// results, the new Offset value should be currentOffset + Limit.
+	Offset uint `url:"offset,omitempty"`
+}
+
+// CreateSlackConnectionWithContext creates a Slack connection.
+func (c *Client) CreateSlackConnectionWithContext(ctx context.Context, slackTeamID string, s SlackConnection) (SlackConnectionObject, error) {
+	d := map[string]SlackConnection{
+		"slack_connection": s,
+	}
+
+	resp, err := c.post(ctx, "/workspaces/"+slackTeamID+"/connections", d, nil)
+	return getSlackConnectionFromResponse(c, resp, err)
+}
+
+// GetSlackConnection gets a Slack connection.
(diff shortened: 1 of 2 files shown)


slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+// https://developer.pagerduty.com/api-reference/c2NoOjExMjA5MzMy-slack-connection.
+type SlackConnectionConfig struct {
+	Events     []string `json:"events"`
+	Urgency    *string  `json:"urgency"`
+	Priorities []string `json:"priorities"`
+}
+
+// SlackConnection is an entity that represents a Slack connections as per the
+// documentation https://developer.pagerduty.com/api-reference/c2NoOjExMjA5MzMy-slack-connection.
+type SlackConnection struct {
```
theckman, line 21: This type seems to be missing the `ID string` field.

slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+	NotificationType string `json:"notification_type"`
+
+	Config SlackConnectionConfig `json:"config"`
+}
+
+// SlackConnectionObject is an API object returned by getter functions.
+type SlackConnectionObject struct {
+	SlackConnection
+	APIObject
+}
```
theckman, line 37: We intend to no longer use embedding in this project. Please list the fields on the type explicitly. See #318 for more details.

That said, I think we should probably just elide this type entirely and put all of the fields in the `SlackConnection` type.

slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+
+// ListSlackConnectionsResponse is an API object returned by the list function.
+type ListSlackConnectionsResponse struct {
+	Connections []SlackConnectionObject `json:"slack_connections"`
+	APIListObject
```
theckman, line 42: Same comment around embedding as well.

slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+	Limit uint `url:"limit,omitempty"`
+
+	// Offset is the pagination parameter that specifies the offset at which to
+	// start pagination results. When trying to request the next page of
+	// results, the new Offset value should be currentOffset + Limit.
+	Offset uint `url:"offset,omitempty"`
+}
+
+// CreateSlackConnectionWithContext creates a Slack connection.
+func (c *Client) CreateSlackConnectionWithContext(ctx context.Context, slackTeamID string, s SlackConnection) (SlackConnectionObject, error) {
```
theckman, line 59: Please remove `WithContext` from the name, since there is not a non-context version.

slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+	d := map[string]SlackConnection{
+		"slack_connection": s,
+	}
+
+	resp, err := c.post(ctx, "/integration-slack/workspaces/"+slackTeamID+"/connections", d, nil)
+	return getSlackConnectionFromResponse(c, resp, err)
+}
+
+// GetSlackConnection gets a Slack connection.
+func (c *Client) GetSlackConnectionWithContext(ctx context.Context, slackTeamID, connectionID string) (SlackConnectionObject, error) {
```
theckman, line 69: Please remove `WithContext` from the name, since there is not a non-context version.

slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+	resp, err := c.get(ctx, "/integration-slack/workspaces/"+slackTeamID+"/connections/"+connectionID)
+	return getSlackConnectionFromResponse(c, resp, err)
+}
+
+// DeleteSlackConnectionWithContext deletes a Slack connection.
+func (c *Client) DeleteSlackConnectionWithContext(ctx context.Context, slackTeamID, connectionID string) error {
```
theckman, line 75: Please remove `WithContext` from the name, since there is not a non-context version.

slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+	_, err := c.delete(ctx, "/integration-slack/workspaces/"+slackTeamID+"/connections/"+connectionID)
+	return err
+}
+
+// UpdateSlackConnectionWithContext updates an existing Slack connection.
+func (c *Client) UpdateSlackConnectionWithContext(ctx context.Context, slackTeamID, connectionID string, s SlackConnection) (SlackConnectionObject, error) {
```
theckman, line 81: Please remove `WithContext` from the name, since there is not a non-context version.

slack_connection.go
```diff
@@ -0,0 +1,128 @@
...
+	d := map[string]SlackConnection{
+		"slack_connection": s,
+	}
+
+	resp, err := c.put(ctx, "/integration-slack/workspaces/"+slackTeamID+"/connections/"+connectionID, d, nil)
+	return getSlackConnectionFromResponse(c, resp, err)
+}
+
+// ListSlackConnectionsWithContext lists Slack connections.
+func (c *Client) ListSlackConnectionsWithContext(ctx context.Context, slackTeamID string, o ListSlackConnectionsOptions) (*ListSlackConnectionsResponse, error) {
```
theckman, line 91: Please remove `WithContext` from the name, since there is not a non-context version. Also, I don't think this should return a pointer type. We're going to be moving away from those when it makes sense (#270).

client.go
```diff
@@ -499,17 +499,25 @@ func (c *Client) prepRequest(req *http.Request, authRequired bool, headers map[s
...
+	}
 
 	dreq := r.Clone(r.Context())
 
-	r.Body = ioutil.NopCloser(bytes.NewReader(data))
-	dreq.Body = ioutil.NopCloser(bytes.NewReader(data))
+	if data != nil {
+		r.Body = ioutil.NopCloser(bytes.NewReader(data))
+		dreq.Body = ioutil.NopCloser(bytes.NewReader(data))
+	}
```
theckman, line 520: Could you please raise this chunk as an isolated PR?

theckman (changes requested):
Thank you for your contribution. I think there are a few changes to be made before we can merge it.

Also, can you please squash the commits of the branch so all of these changes are within a single commit? It helps keep the history of the primary branch more clear.

RangelReale:
any follow up on this one?

PR #436, opened 2022-04-06
Fix for `go install`

### LINKED ISSUE #435: Update install instructions for "go install"
Since `go install` is now the recommended approach https://tip.golang.org/doc/go1.16#modules, steps/README should be updated accordingly.

Currently `go install github.com/PagerDuty/go-pagerduty/command@latest` works, but results in a binary named "command". I would like to propose nesting the contents of command into another `pd` subfolder such that `go install github.com/PagerDuty/go-pagerduty/command/pd@latest` would work as expected.

mjlshen:
By moving the contents of command into a `command/pd` subdirectory, go install can now be used to install a binary.

I updated the README to demonstrate an example for latest, since the instructions won't work for v1.5.0, but maybe when the next version is released the example in the README can be pinned to a version?

Fixes #435


mjlshen:
@theckman Not sure if you have cycles to review this - totally ok if you don't, but if I should ping someone else please let me know, thanks!

mjlshen:
Tagging @stmcallister - noticed that you've reviewed other PRs

theckman:
I think we need to figure out what we want to do with the command line tool. There's been a discussion in the past about this being removed from the project. I wonder if it makes sense to ask a PagerDuty employee to create a new repo, and migrate the code there so that it's go-getable. 

@stmcallister would that be something we could make happen?

sodabrew:
Subscribe. watching for Go 1.19 compatibility to `go install...` the cli.

devth:
Any updates on the status of the command line tool? I'd love to start using it.

PR #437, opened 2022-04-12
Ars/webhook updates

arussellsaw:
having non-pointer non-omitempty fields meant we were sending invalid params for updates, this means we only send the fields we set.

### DIFF
--- a/webhook.go
+++ b/webhook.go
@@ -101,10 +101,11 @@ type WebhookResponse struct {
 }
 
 type UpdateWebhookOptions struct {
-	Active      bool     `json:"active"`
-	Description string   `json:"description"`
-	Events      []string `json:"events"`
-	Filter      Filter   `json:"filter"`
+	// pointer fields here are used to allow us to omit certain fields when updating
+	Active      *bool    `json:"active,omitempty"`
+	Description *string  `json:"description,omitempty"`
+	Events      []string `json:"events,omitempty"`
+	Filter      *Filter  `json:"filter,omitempty"`
 }
 
 // CreateWebhookWithContext creates a new webhook.
--- a/webhook.go
+++ b/webhook.go
@@ -66,6 +66,9 @@ type DeliveryMethod struct {
 	Type                string   `json:"type"`
 	CustomHeaders       []string `json:"custom_headers"`
 	TemporarilyDisabled bool     `json:"temporarily_disabled"`
+	ID                  string   `json:"id"`
+	Secret              string   `json:"secret"`
+	ExtensionID         string   `json:"extension_id"`
 }
 
 type Filter struct {
--- a/webhook.go
+++ b/webhook.go
@@ -1,6 +1,7 @@
 package pagerduty
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"time"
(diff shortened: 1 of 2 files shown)


webhook.go
```diff
@@ -59,3 +60,107 @@ func DecodeWebhook(r io.Reader) (*WebhookPayloadMessages, error) {
...
+	var ii WebhookResponse
+	if err = c.decodeJSON(resp, &ii); err != nil {
+		return nil, err
+	}
+
+	return &ii.WebhookSubscription, nil
+}
+
+// GetWebhookWithContext returns information about a specific webhook by ID
+func (c *Client) GetWebhookWithContext(ctx context.Context, id string) (*WebhookSubscription, error) {
```
theckman, line 152: Same thoughts as the comment on the `Create` method.

webhook_test.go
```diff
@@ -39,3 +44,201 @@ func TestWebhook_DecodeWebhook(t *testing.T) {
...
+		},
+	}
+
+	mux.HandleFunc("/webhook_subscriptions", func(w http.ResponseWriter, r *http.Request) {
+		testMethod(t, r, "POST")
+
+		body, err := ioutil.ReadAll(r.Body)
+		testErrCheck(t, "ioutil.ReadAll", "", err)
+
+		fmt.Println(string(body))
```
theckman, line 72: Not sure this is needed.

webhook_test.go
```diff
@@ -39,3 +44,201 @@ func TestWebhook_DecodeWebhook(t *testing.T) {
...
+			Type: "http_delivery_method",
+		},
+		Description: "my new webhook",
+		Events:      []string{"incident.triggered"},
+		Filter: Filter{
+			Type: "account_reference",
+		},
+	}
+
+	if err != nil {
```
theckman, line 128: I'd generally use `testErrCheck` instead, to get the standard error output that you'd see on other tests.

webhook_test.go
```diff
@@ -39,3 +44,201 @@ func TestWebhook_DecodeWebhook(t *testing.T) {
...
+			Type: "http_delivery_method",
+		},
+		Description: "my new webhook",
+		Events:      []string{"incident.triggered", "incident.delegated"},
+		Filter: Filter{
+			Type: "account_reference",
+		},
+	}
+
+	if err != nil {
```
theckman, line 187: Same thought on `testErrCheck` here.

webhook_test.go
```diff
@@ -39,3 +44,201 @@ func TestWebhook_DecodeWebhook(t *testing.T) {
...
+			Type: "http_delivery_method",
+		},
+		Description: "my new webhook",
+		Events:      []string{"incident.triggered"},
+		Filter: Filter{
+			Type: "account_reference",
+		},
+	}
+
+	if err != nil {
```
theckman, line 240: Same thought on `testErrCheck` here.

theckman (changes requested):
Thank you for the contribution, I do have some changes I think should be made before we merge this in.

arussellsaw:
Oh wow I didn't realise I raised this 🙈 I think I misclicked when trying to PR against our internal fork. This would be nice to have upstream anyway, but sorry for accidentally asking for your review when it wasn't as polished as it should be. I'll see if I can find some time this week to tidy up

PR #434, opened 2022-04-06
Update golang.org/x/sys to fix macOS builds

mjlshen:
Fixes #433

Build completes successfully afterwards:
```bash
❯ make build
go get
go mod verify
all modules verified
go mod vendor
go build -mod=vendor -o pd ./command

❯ ./pd
Usage: pd [--version] [--help] <command> [<args>]

Available commands are:
    ability
    addon
    analytics
    escalation-policy
    event-v2
    incident
    log-entry
    maintenance-window
    notification
    oncall
    schedule
    service
    team
    user
    vendor
```

PR #442, opened 2022-04-22
Fix the formatting of the ResponderRequest input and output

CerealBoy:
The [documentation](https://developer.pagerduty.com/api-reference/b3A6Mjc0ODE1MQ-create-a-responder-request-for-an-incident) has 2 different ways to structure this code, where the doc and the example are different. Previously the code here had worked with the doc formatting, however this is currently not working as intended.

The main change here is within the ``responder_request_targets`` portion of the payload, where the array content is an object that then contains a key ``responder_request_target`` with the actual object, rather than just this object itself. i.e. It's nested one key deeper in the tree. Originally I'd thought this was different for the input and the output, turns out the only difference is that the ``incident_responders`` is not provided in the input, and populated in the output.

This PR also includes the 3 line change to update the ``go.*`` files as this was tested on an M1 mac, so a newer version of the ``golang.org/x/sys`` library is needed that supports ``arm64``.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,5 +8,6 @@ require (
 	github.com/mitchellh/cli v1.0.0
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/sirupsen/logrus v1.4.2
+	golang.org/x/sys v0.0.0-20220412211240-33da011f77ad // indirect
 	gopkg.in/yaml.v2 v2.2.2
 )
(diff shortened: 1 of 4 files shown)


ChuckCrawford:
Apologies for not getting to this earlier but it looks like we have resolved these issues as part of https://github.com/PagerDuty/go-pagerduty/pull/452 and https://github.com/PagerDuty/go-pagerduty/pull/434.

Feel free to reopen this if there are still issues in this area.

PR #443, opened 2022-04-23
Fix potential panic when debug request captures are enabled

theckman:
This fixes a panic that would occur if we attempted to duplicate a request that
has a nil Body, which would be any `GET` request. This also adds a regression
test.


theckman:
@stmcallister the build looks to have passed, but I don't see the status being reflected here. 🤔 https://app.circleci.com/pipelines/github/PagerDuty/go-pagerduty?branch=fix_captured_request_panic&filter=all

stmcallister (approved):
Nice! And, thank you for adding the test! 👍

PR #444, opened 2022-04-23
Backport fix for panic when debug request captures are enabled

theckman:
This fixes a panic that would occur if we attempted to duplicate a request that
has a nil Body, which would be any `GET` request. This also adds a regression
test.

### DIFF
--- a/client.go
+++ b/client.go
@@ -497,17 +497,19 @@ func (c *Client) prepRequest(req *http.Request, authRequired bool, headers map[s
 }
 
 func dupeRequest(r *http.Request) (*http.Request, error) {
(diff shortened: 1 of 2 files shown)


theckman:
@stmcallister this one too; build passed but status wasn't updated: https://app.circleci.com/pipelines/github/PagerDuty/go-pagerduty?branch=fix_debug_panic_release-1.5.x&filter=all

theckman:
I'm going to merge this one in as well, since I can merge things into `release-1.5.x`. I've confirmed the fix works as intended and have included tests. I do need a review on #443 to make sure the same fix is included in `v1.6.0`.

PR #448, opened 2022-05-01
Fix parsing bug for AlertGroupingParameters

### LINKED ISSUE #438: Chaining GetService and UpdateService fails when AlertGroupingParameters is omitted
```go
service, err := pd.GetService(serviceID, nil)
if err != nil {
    return err
}

service.status = "disabled"

if _, err = pd.UpdateService(*service); err != nil {
    return err
}
```

Results in the following when the following error when `AlertGroupingParameters` is unset/omitted by Go:
```
"error":"HTTP response failed with status code 400, message: Invalid Input Provided (code: 2001): Alert grouping parameters is invalid."
```

The JSON response from GetService looks like:
```json
"alert_grouping_parameters": {
      "type": null,
      "config": null
    },
```

but the resulting struct ends up like, which the PagerDuty API doesn't seem to like:
```go
pd.AlertGroupingParameters{
	Type:   "",
	Config: pd.AlertGroupParamsConfig{
		Timeout:   0,
		Aggregate: "",
		Fields:    nil,
	},
}
```

mjlshen:
The PagerDuty API is happy when only `Type` is set to `omitempty`, but I think it makes sense to do the same for `AlertGroupParamsConfig` as well as change it to be a pointer so that it gets a nil value instead of the zero value for all the config values in case the api checks those values in the future.

```go
func main() {
	client := pagerduty.NewClient(authToken)
	svc, err := client.GetServiceWithContext(context.TODO(), serviceId, nil)
	if err != nil {
		panic(err)
	}

        // Whether there's no value from GetService or if it's set
	svc.AlertGroupingParameters.Type = ""

	_, err := client.UpdateServiceWithContext(context.TODO(), *svc)
	if err != nil {
                 // Results in panic: HTTP response failed with status code 400,
                 // message: Invalid Input Provided (code: 2001): Alert grouping parameters is invalid.
		panic(err)
	}
}
```

Fixes #438

### DIFF
--- a/service.go
+++ b/service.go
@@ -99,10 +99,10 @@ type Service struct {
 	Addons                  []Addon                  `json:"addons,omitempty"`
 }
 
-// AlertGroupingParameters defines how alerts on the servicewill be automatically grouped into incidents
+// AlertGroupingParameters defines how alerts on the service will be automatically grouped into incidents
 type AlertGroupingParameters struct {
-	Type   string                 `json:"type"`
-	Config AlertGroupParamsConfig `json:"config"`
+	Type   string                  `json:"type,omitempty"`
+	Config *AlertGroupParamsConfig `json:"config,omitempty"`
 }
 
 // AlertGroupParamsConfig is the config object on alert_grouping_parameters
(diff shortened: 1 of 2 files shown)


mjlshen:
@theckman @ChuckCrawford not sure if any of you have cycles to review this? Thanks!

ChuckCrawford (approved):

ChuckCrawford:
Thank you for your help with this @mjlshen .

Issue #451, opened 2022-07-20
ResponderRequest regression - API method is no longer functional


jwestbury:
Looks like ResponderRequest was fixed in #328, but a regression was introduced in #389 which broke this method again. I'm seeing the following request body from the SDK, which results in a 2100 (Not Found) response:

```
{
    "message": "Testing",
    "requester_id": "P1P12P1",
    "responder_request_targets": [
        {
            "id": "P1P12P1",
            "type": "user_reference"
        }
    ]
}
```

Meanwhile, the API explorer produces the following request body, which works with an identical issue ID:

```
{
  "message": "Testing",
  "requester_id": "P1P12P1",
  "responder_request_targets": [
    {
      "responder_request_target": {
        "id": "P1P12P1",
        "type": "user_reference"
      }
    }
  ]
}
```

Any chance of getting the "correct" behavior reintroduced?

CerealBoy:
Would this be resolved by the (now outdated) #442 ?

ChuckCrawford:
This should now be resolved by https://github.com/PagerDuty/go-pagerduty/pull/452 and included in the eventual 1.6 release. 

Thank you all for your patience. Please let us know if anyone runs into additional issues here.

PR #452, opened 2022-07-29
Fix ResponderRequest regression

ehlerorngard:
This provides a fix for issue #451 where _ResponderRequest_ has been broken (requests are returned a 2100 _Not Found_). 

The PR also omits `incident_responders` from the request when empty, as it's not required/useful.

I fixed the related test in `incident_test.go` and tested successfully both with that test and my own test which actually sent the successful POSTs to the `/incidents/{id}/responder_requests` endpoint.

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -657,7 +657,12 @@ type ResponderRequestResponse struct {
 // ResponderRequestTarget specifies an individual target for the responder request.
 type ResponderRequestTarget struct {
 	APIObject
-	Responders IncidentResponders `json:"incident_responders"`
+	Responders IncidentResponders `json:"incident_responders,omitempty"`
+}
+
+// ResponderRequestTargetWrapper is a wrapper for a ResponderRequestTarget.
+type ResponderRequestTargetWrapper struct {
+	Target ResponderRequestTarget `json:"responder_request_target"`
 }
 
 // ResponderRequestOptions defines the input options for the Create Responder function.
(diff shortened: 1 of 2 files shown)


incident.go
```diff
@@ -657,15 +657,20 @@ type ResponderRequestResponse struct {
 // ResponderRequestTarget specifies an individual target for the responder request.
 type ResponderRequestTarget struct {
 	APIObject
-	Responders IncidentResponders `json:"incident_responders"`
+	Responders IncidentResponders `json:"incident_responders,omitempty"`
+}
+
+// ResponderRequestTargets is a wrapper for a ResponderRequestTarget.
+type ResponderRequestTargets struct {
```
ChuckCrawford, line 664: I know this is a little strange but could we call this something like `ResponderRequestTargetWrapper`? Calling it `ResponderRequestTargets` [plural] is a little strange because it isn't actually a list or collection of any kind...

ChuckCrawford (commented):
@ehlerorngard : would you mind having a look at the one small comment I left? Then this should be good to go.

incident.go
ehlerorngard: @ChuckCrawford Thanks so much for taking a look at this!  And I totally agree - `ResponderRequestTargetWrapper` is both clearer and more accurate (I had noticed that false pluralization as well, but I chickened out of changing the name and just left it as it was named in its last working state).  Edit made, and squashed the commits to keep the history tidier.

ChuckCrawford (approved):

PR #454, opened 2022-08-04
Prevent leak on error responses

ChuckCrawford:
Reverts https://github.com/PagerDuty/go-pagerduty/pull/453 and instead applies the fix in a common location.

### DIFF
--- a/client.go
+++ b/client.go
@@ -590,6 +590,8 @@ func (c *Client) checkResponse(resp *http.Response, err error) (*http.Response, 
 func (c *Client) getErrorFromResponse(resp *http.Response) APIError {
 	// check whether the error response is declared as JSON
 	if !strings.HasPrefix(resp.Header.Get("Content-Type"), "application/json") {
+		defer resp.Body.Close()
+
 		aerr := APIError{
 			StatusCode: resp.StatusCode,
 			message:    fmt.Sprintf("HTTP response with status code %d does not contain Content-Type: application/json", resp.StatusCode),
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -590,6 +590,8 @@ func (c *Client) checkResponse(resp *http.Response, err error) (*http.Response,
 func (c *Client) getErrorFromResponse(resp *http.Response) APIError {
 	// check whether the error response is declared as JSON
 	if !strings.HasPrefix(resp.Header.Get("Content-Type"), "application/json") {
+		defer func() { _ = resp.Body.Close() }()
```
acloudgopher-pd, line 593: nit: i suppose this is the same as defer resp.Body.Close() if we don't need to check the error

acloudgopher-pd (approved):
Looks ok to me all tests pass, but I'm a bit curious why only check for non json content

ChuckCrawford:
> Looks ok to me all tests pass, but I'm a bit curious why only check for non json content

Ah good question @acloudgopher-pd. I should have left some comments describing my thought process. So all of the requests eventually call `c.checkResponse(resp, err)` which has some common handling of non-2xx status codes: https://github.com/PagerDuty/go-pagerduty/blob/8895770d9489ca6ccb18f76d236bb52a432df68d/client.go#L583-L585

inside of `getErrorFromResponse`, the call to `decodeJSON` has handling to close the response body: https://github.com/PagerDuty/go-pagerduty/blob/8895770d9489ca6ccb18f76d236bb52a432df68d/client.go#L560-L566

so it looked like the only place this was missing was for the scenario where a non-JSON response was returned...

rafusel (approved):
This looks like a better approach to me 👍

PR #460, opened 2022-08-16
Update to Go 1.19

mjlshen:
None of the dependencies' versions were updated, just updating Go

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/PagerDuty/go-pagerduty
 
-go 1.14
+go 1.19
 
 require (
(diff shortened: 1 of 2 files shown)


ChuckCrawford:
Hey @mjlshen: thank you for your contribution here!

While it is probably time for a Golang upgrade, I am a little hesitant to directly upgrade to 1.19 without giving clients a bit more time to upgrade themselves.

What do you think about https://github.com/PagerDuty/go-pagerduty/pull/466 to bump the version to 1.17?

mjlshen:
Sounds good to me to do this more incrementally! I'll close this PR

PR #422, opened 2022-02-10
Adding GET API for Analytics raw incident data for a given incident

eepurichandra:
# Changes
* Implementation corresponds to `/analytics/raw/incidents/id`  pagerduty API.     
https://developer.pagerduty.com/api-reference/b3A6Mjc0ODExMw-get-raw-data-single-incident

PR #481, opened 2023-05-04
Add raw incidents metrics

MaggotMouth:
Implement endpoints for single and multiple raw incident metrics

https://developer.pagerduty.com/api-reference/b3A6Mjc0ODExMw-get-raw-data-single-incident

https://developer.pagerduty.com/api-reference/c2d493e995071-get-raw-data-multiple-incidents

This was previously done in PRs #347 and #422 which have gone stale.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -17,6 +17,8 @@ type AnalyticsRequest struct {
 	TimeZone      string           `json:"time_zone,omitempty"`
 }
 
+// RawDataRequest represents the request to be sent to PagerDuty when you want
+// raw analytics.
 type RawDataRequest struct {
 	Filters       *AnalyticsFilter `json:"filters,omitempty"`
 	StartingAfter string           `json:"starting_after,omitempty"`
@@ -35,6 +37,7 @@ type AnalyticsResponse struct {
 	TimeZone      string           `json:"time_zone,omitempty"`
 }
 
+// RawDataResponse represents the response from the PagerDuty API.
 type RawDataResponse struct {
 	First    string           `json:"first,omitempty"`
 	Last     string           `json:"last,omitempty"`
@@ -60,7 +63,7 @@ type AnalyticsFilter struct {
 	PriorityNames  []string `json:"priority_names,omitempty"`
 }
 
-// AnalyticsData represents the structure of the analytics we have available.
+// AnalyticsData represents the structure of the aggregated analytics we have available.
 type AnalyticsData struct {
 	ServiceID                      string  `json:"service_id,omitempty"`
 	ServiceName                    string  `json:"service_name,omitempty"`
@@ -85,6 +88,7 @@ type AnalyticsData struct {
 	RangeStart                     string  `json:"range_start,omitempty"`
 }
 
+// RawData represents the structure of the raw analytics we have available.
 type RawData struct {
 	AssignmentCount           int    `json:"assignment_count,omitempty"`
 	BusinessHourInterruptions int    `json:"business_hour_interruptions,omitempty"`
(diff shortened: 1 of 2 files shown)


krav:
@ChuckCrawford @theckman Hi! Can we get this reviewed? Been looking to use this endpoint. Thanks!

analytics.go
```diff
@@ -94,3 +151,45 @@ func (c *Client) getAggregatedData(ctx context.Context, analytics AnalyticsReque
 
 	return analyticsResponse, nil
 }
+
+// GetRawDataSingleIncident gets the raw analytics for the requested incident.
+func (c *Client) GetRawDataSingleIncident(ctx context.Context, id string) (RawData, error) {
+	h := map[string]string{
+		"X-EARLY-ACCESS": "analytics-v2",
```
ChuckCrawford, line 158: They probably were at the time you put up the PR but these APIs are no longer "Early Access"

analytics.go
```diff
@@ -94,3 +151,45 @@ func (c *Client) getAggregatedData(ctx context.Context, analytics AnalyticsReque
 
 	return analyticsResponse, nil
 }
+
+// GetRawDataSingleIncident gets the raw analytics for the requested incident.
+func (c *Client) GetRawDataSingleIncident(ctx context.Context, id string) (RawData, error) {
```
ChuckCrawford, line 156: Could we update the signatures to return a pointer to the result? e.g. `func (c *Client) GetRawDataSingleIncident(ctx context.Context, id string) (*RawData, error)`

ChuckCrawford, line 156: A few nitpicky comments about the function names: could we name these two functions after their [OpenAPI Operation Ids](https://github.com/PagerDuty/api-schema/blob/55f11455b56cd1a829e0a949ebba744087c75172/reference/REST/openapiv3.json#L18401): `getAnalyticsIncidentsById` and `getAnalyticsIncidents`?

analytics.go
```diff
@@ -61,6 +88,36 @@ type AnalyticsData struct {
 	RangeStart                     string  `json:"range_start,omitempty"`
 }
 
+// RawData represents the structure of the raw analytics we have available.
+type RawData struct {
```
ChuckCrawford, line 92: Could we name this `AnalyticsRawIncident` ?

analytics.go
```diff
@@ -23,6 +37,19 @@ type AnalyticsResponse struct {
 	TimeZone      string           `json:"time_zone,omitempty"`
 }
 
+// RawDataResponse represents the response from the PagerDuty API.
+type RawDataResponse struct {
```
ChuckCrawford, line 41: This would be `AnalyticsRawIncidentsResponse`

analytics.go
```diff
@@ -15,6 +17,18 @@ type AnalyticsRequest struct {
 	TimeZone      string           `json:"time_zone,omitempty"`
 }
 
+// RawDataRequest represents the request to be sent to PagerDuty when you want
+// raw analytics.
+type RawDataRequest struct {
```
ChuckCrawford, line 22: This would be `AnalyticsRawIncidentsRequest`

ChuckCrawford (changes requested):
Thank you for the PR! If would wouldn't mind making a few changes we can get this merged...

MaggotMouth:
Thanks for the review @ChuckCrawford - I've made the changes as requested and updated the associated tests.

ChuckCrawford (approved):
Thank you for your contribution @MaggotMouth.

Tyrael:
@ChuckCrawford is there any ETA for v1.9.0? this was merged more than a year ago but there were no releases since.

PR #482, opened 2023-05-09
Add resolved_at and updated_at fields to Incident

surik:
As per documentation: https://developer.pagerduty.com/api-reference/9d0b4b12e36f9-list-incidents


dnwe:
Commit and PR title should be `Add resolved_at and updated_at fields to Incident` as `created_at` already exists. Otherwise change is correct and working well via a `replace { .. }` directive. Would be good to get this fix into the next version release

surik:
@dnwe thank you for pointing out. This is fixed.

PR #487, opened 2023-06-29
Add support for incident custom fields

wishperera:
## Why

- support for listing/setting custom fields for an incident
- Reference: 
      - https://developer.pagerduty.com/api-reference/85a33d421d50b-get-custom-field-values
      - https://developer.pagerduty.com/api-reference/0f6094f852517-update-custom-field-values

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -904,7 +904,7 @@ func (c *Client) UpdateIncidentCustomFieldsWithContext(ctx context.Context, id s
 		"custom_fields": fields,
 	}
 
-	resp, err := c.post(ctx, "/incidents/"+id+"/custom_fields/values", d, nil)
+	resp, err := c.put(ctx, "/incidents/"+id+"/custom_fields/values", d, nil)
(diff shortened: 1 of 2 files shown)


thomas-maurice:
Yesssss please !

incident.go
```diff
@@ -874,3 +874,59 @@ func (c *Client) RemoveIncidentNotificationSubscribersWithContext(ctx context.Co
...
+
+type IncidentCustomField struct {
+	ID          string `json:"id"`
+	Type        string `json:"type"`
+	Name        string `json:"name"`
+	DisplayName string `json:"display_name"`
+	Description string `json:"description"`
+	DataType    string `json:"data_type"`
+	FieldType   string `json:"field_type"`
+	Value       string `json:"value"`
```
thomas-maurice, line 891: This won't work, because sometimes values can be slices for example in the case of a `multi_value_fixed` field type

PR #488, opened 2023-07-02
Refactor nil *string initialization

typeid:
**What's the purpose of this PR?** 

- This PR removes the `strPtr` helper function, as it is not necessary and was missleading.

**Why?**

- The `strPtr` function was used to initialize a string and get its pointer in a single line.  As this was only used for initializing empty strings and returning their pointer, it's equivalent to `new(string)`. For unit tests, it makes sense to reference the same values for the created json and the unmarshalled output.
- The `strPtr` function was missleading because it took a string by copy instead of reference, meaning that it would not return the pointer to the passed string, but to the copy.

### DIFF
--- a/service_integration.go
+++ b/service_integration.go
@@ -241,27 +241,25 @@ func (i *IntegrationEmailFilterRule) UnmarshalJSON(b []byte) error {
 
 	// if the *string is nil, set it to a *string with value ""
 	if ief.SubjectRegex == nil {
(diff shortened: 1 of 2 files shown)


ChuckCrawford:
Makes good sense. Thank you for the contribution @typeid.

ChuckCrawford (approved):

PR #490, opened 2023-07-24
Added support for auto_pause_notifications_parameters

darrendao:
Updated Service struct to support auto_pause_notifications_parameters as per https://developer.pagerduty.com/api-reference/165ad96a22ffd-get-a-service.
 I'm guessing this is a newer feature that got released but never made its way into this go lib code.

### DIFF
--- a/service.go
+++ b/service.go
@@ -78,25 +78,32 @@ type ServiceRuleActions struct {
 // Service represents something you monitor (like a web service, email service, or database service).
 type Service struct {
 	APIObject
-	Name                    string                   `json:"name,omitempty"`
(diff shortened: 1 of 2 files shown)


darrendao:
@ChuckCrawford is this something you can help review/approve/merge?

darrendao:
@jolexa @joelcrocker @helderberto @gempesaw @gsreynolds @cmrberry Is this something any of you can help get moving?

ChuckCrawford (approved):
> I'm guessing this is a newer feature that got released but never made its way into this go lib code.

You are correct @darrendao . Thank you for contributing this and for following up with us!

Issue #471, opened 2022-12-07
Missing support for raw analytics


ztzxt:
Hi, as fas as I could see Go client doesn't support those two endpoints:

> https://developer.pagerduty.com/api-reference/c2d493e995071-get-raw-data-multiple-incidents

> https://developer.pagerduty.com/api-reference/328d94baeaa0e-get-raw-data-single-incident

Are those endpoints supported in client library? I would love to try implementing them if they don't exist yet.

ChuckCrawford:
I would guess they are not yet implemented since they are in Early Access. I am not 100% sure how we have handled Early Access APIs in our client library but I would welcome the PR if you wanted to attempt adding the support...

ChuckCrawford:
This should now be resolved with: https://github.com/PagerDuty/go-pagerduty/pull/481

PR #491, opened 2023-07-25
wip: Implement RawAnalytics endpoint

ArthurSens:
Related to #471 

This PR aims to implement https://developer.pagerduty.com/api-reference/4c464d0f1cabc-get-raw-responses-from-a-single-incident

I'm opening this early as a draft since I'm not sure if the approach I took with the early access header is the best one. Would love some feedback from the maintainers :)


ChuckCrawford:
@ArthurSens: thank you for this PR. This should now be resolved by: https://github.com/PagerDuty/go-pagerduty/pull/481

PR #493, opened 2023-08-08
Fix ResponderRequest unmarshalling of IncidentResponders

allyjweir:
This PR fixes an unmarshalling bug that was found while trying to use the `ResponderRequestWithContext` method on the client.

The response that the API returns is subtly different to the one which the client library is expecting. Within the `responder_request_target` object, the client library expects a key called `incident_responders`. However when interacting with the API directly, this key is actually `incidents_responders`. Note the additional `s` which pluralises 'incident'.

## Impact

This meant that whenever you made a responder request via the client library, it would succeed but the response would only be partial. That meant you could not retrieve any of the information about the responders which your request had added to an incident.

## A note on API documentation

Please note that there is an inconsistency in the [PagerDuty API documentation for this endpoint](https://developer.pagerduty.com/api-reference/6850764326ee2-create-a-responder-request-for-an-incident).

- In the written documentation it refers to: `incident_responders`
- In the 'response example', it correctly refers to: `incidents_responders`

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -659,7 +659,7 @@ type ResponderRequestResponse struct {
 // ResponderRequestTarget specifies an individual target for the responder request.
 type ResponderRequestTarget struct {
 	APIObject
-	Responders IncidentResponders `json:"incident_responders,omitempty"`
+	Responders []IncidentResponders `json:"incidents_responders,omitempty"`
 }
 
 // ResponderRequestTargetWrapper is a wrapper for a ResponderRequestTarget.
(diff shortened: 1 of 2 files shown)


incident.go
```diff
@@ -659,7 +659,7 @@ type ResponderRequestResponse struct {
 // ResponderRequestTarget specifies an individual target for the responder request.
 type ResponderRequestTarget struct {
 	APIObject
-	Responders IncidentResponders `json:"incident_responders,omitempty"`
+	Responders []IncidentResponders `json:"incidents_responders,omitempty"`
```
allyjweir, line 662: I did consider if the `IncidentResponders` type should be renamed `IncidentResponder` but decided to keep this PR to the minimum diff and not to break the interface

ChuckCrawford:
Thanks for submitting this @allyjweir. Just following up on something internally and then we will get this merged.

ChuckCrawford (approved):

Issue #480, opened 2023-05-01
Please add support for the license APIs


mk23:
* https://developer.pagerduty.com/api-reference/10230a10c519e-list-license-allocations
* https://developer.pagerduty.com/api-reference/4c10cb38f7381-list-licenses

gerardocorea:
Wouldn't mind take a stab at this. Is there any specific process to confirm this would be approved if the implementation is accepted?

ChuckCrawford:
@gerardocorea : if you put up a PR we will do our best to get it reviewed and merged. You can `@me` when you get it done...

gerardocorea:
@ChuckCrawford Thanks for the response, what might be the best way to go about testing this if someone doesn't have access to an enterprise account with licenses.

gerardocorea:
@ChuckCrawford opened a PR for this request.

PR #494, opened 2023-09-02
adding support for license and licenses allocated based operations #480

gerardocorea:
Following a similar approach to listing existing pagerduty resources from the pagerduty api, this PR adds the ability to list available licenses as well as licenses allocated. #480

### DIFF
--- a/license.go
+++ b/license.go
@@ -1,0 +1,84 @@
+package pagerduty
+
+import (
+	"context"
+
+	"github.com/google/go-querystring/query"
+)
+
+type License struct {
+	APIObject
+	Name                 string   `json:"name"`
+	Description          string   `json:"description"`
+	ValidRoles           []string `json:"valid_roles"`
+	RoleGroup            string   `json:"role_group"`
+	Summary              string   `json:"summary"`
+	CurrentValue         int      `json:"current_value"`
+	AllocationsAvailable int      `json:"allocations_available"`
+}
+
+type LicenseAllocated struct {
+	APIObject
+	Name        string   `json:"name"`
+	Description string   `json:"description"`
+	ValidRoles  []string `json:"valid_roles"`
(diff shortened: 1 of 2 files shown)


license.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (c *Client) ListLicenses() (*ListLicensesResponse, error) {
+	return c.ListLicensesWithContext()
+}
+
+func (c *Client) ListLicenseAllocations(o ListLicenseAllocationsOptions) (*ListLicenseAllocationsResponse, error) {
+	return c.ListLicenseAllocationsWithContext(o)
+}
+
+func (c *Client) ListLicensesWithContext() (*ListLicensesResponse, error) {
```
ChuckCrawford, line 57: The *WithContext() methods should accept a `context.Context` parameter. The newly added Event Orchestration APIs should give you an example of this: https://github.com/PagerDuty/go-pagerduty/blob/main/event_orchestration.go#L73-L90.

ChuckCrawford (changes requested):
Thank you for putting up this PR. If we can get the method signatures reworked to properly handle the request context [comment here](https://github.com/PagerDuty/go-pagerduty/pull/494/files#r1329067020) we should be looking alright.

license.go
ChuckCrawford: Could we also remove the non-context methods since that is now a deprecated pattern?

gerardocorea:
> Thank you for putting up this PR. If we can get the method signatures reworked to properly handle the request context [comment here](https://github.com/PagerDuty/go-pagerduty/pull/494/files#r1329067020) we should be looking alright.

Appreciate the feedback, made the requested changes.

ChuckCrawford (approved):
Thank you for your contribution @gerardocorea.

PR #501, opened 2023-10-26
[CSGI-1873] Support Scoped OAuth App authentication

imjaroiswebdev:
This update is meant to bring support for [Scoped OAuth App authentication](https://developer.pagerduty.com/docs/e518101fde5f3-obtaining-an-app-o-auth-token) capabilities to Go PagerDuty library.

Therefore, the new Client constructor method `NewScopedOAuthAppClient` was introduced to enable this authentication method.

On top of that, due to **Access Token** for Scoped OAuth should be re-used till its expiration, then persistency of obtained *Access Token* had to be included; Since the **CLI** client was already using the file `~/.pd.yml`, this same file is the default location for persisting *Access Token*.

However, this location can be changed to any other valid location and file name at client configuration. Like is described in updated `README` regarding this same topic.


ChuckCrawford (commented):
@imjaroiswebdev: I am still playing around with alternatives here but I wanted to be sure you had a bit of feedback.

I am a bit uncomfortable adding in client code that is reading from disk on every HTTP request [and writing to disk by default as well]. In scenarios where this API client is used in something like a [12-factor-app](https://12factor.net/) this seems undesirable.

What do you think about focusing this first PR on production-ready support for Scoped OAuth when the library is used as an API client [and holding off on CLI support for the moment].

For that scenario it feels like we would want the client library to be able to:
* Be able to concurrently make requests from multiple goroutines
* Obtain new access token(s) when the current token(s) expire

We might also want to consider using: https://pkg.go.dev/golang.org/x/oauth2/clientcredentials.

client.go
```diff
@@ -346,6 +363,34 @@ func WithOAuth() ClientOptions {
 	}
 }
 
+func WithScopedOAuthApp(clientId string, clientSecret string, scopes []string) ClientOptions {
```
imjaroiswebdev, line 366: Since support for User Scoped OAuth token will be also added and this is a public API, I decided to distinguish it from the eventual implementation of `WithScopedOAuthUser` in advance. Due to `WithScopedOAuthUser` will use `oauth2.Config` and this `oauth2.Credentials`. I did the same distinction with `WithScopedOAuthAppTokenSource`.

README.md
```diff
@@ -57,13 +57,14 @@ upgrade for some reason, the `v1.4.x` line of releases should still work. At the
 time of writing `v1.4.3` was the latest, and we intend to backport any critical
 fixes for the time being.
 
-#### Example Usage
+#### Example Usage with [API Token authentication](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTUx-authentication)
```
ChuckCrawford, line 60: nit: Authentication

client.go
```diff
@@ -346,6 +363,34 @@ func WithOAuth() ClientOptions {
...
+	ts := baseTokenSource(context.Background(), clientId, clientSecret, scopes)
```
ChuckCrawford, line 367: We should probably remove `context.Background()` and pass it in as a parameter to `WithScopedOAuthApp`

token_source.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("failed to read file %s: %w", c.configFilePath, err)
+	}
+	pt := persistedToken{}
+	if err := json.Unmarshal(data, &pt); err != nil {
+		return nil, fmt.Errorf("failed to decode content of %s: %w", c.configFilePath, err)
+	}
+
+	t := pt.Token
+
+	// [*] TODO: Persisted token is already expired in config file.
```
ChuckCrawford, line 63: Are these still TODOs? Are we wanting to merge this before those TODOs are complete?

token_source.go
```diff
@@ -0,0 +1,133 @@
...
+		configFilePath: configFilePath,
+		clientId:       clientId,
+		scopes:         scopes,
+	}
+
+	return oauth2.ReuseTokenSource(nil, fts)
+}
+
+func (c *fileTokenSource) loadToken() (*oauth2.Token, error) {
+	log.Printf("[FTS] Loading token from file\n")
```
ChuckCrawford, line 45: Do we still need these log statements? Today only the CLI commands log anything.

token_source.go
```diff
@@ -0,0 +1,133 @@
+package pagerduty
```
ChuckCrawford, line 1: nit: maybe rename this to `file_token_source.go` ?

client.go
```diff
@@ -78,9 +83,10 @@ type APIDetails struct {
 // occurs. This includes messages that should hopefully provide useful context
 // to the end user.
 type APIErrorObject struct {
-	Code    int      `json:"code,omitempty"`
-	Message string   `json:"message,omitempty"`
-	Errors  []string `json:"errors,omitempty"`
+	Code           int      `json:"code,omitempty"`
+	Message        string   `json:"message,omitempty"`
+	Errors         []string `json:"errors,omitempty"`
+	RequiredScopes string   `json:"required_scopes,omitempty"`
```
ChuckCrawford, line 89: This still feels a little weird to me since it is not part of our standard API Error schema. Let me follow up on this one and get back to you.

imjaroiswebdev: Let's handle required scopes through `errors` array as the standard way.

README.md
```diff
@@ -57,13 +57,14 @@ upgrade for some reason, the `v1.4.x` line of releases should still work. At the
...
 
-#### Example Usage
+#### Example Usage with [API Token Authentication](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTUx-authentication)
 
 ```go
 package main
 
 import (
 	"fmt"
+    "context"
```
ChuckCrawford, line 67: nit: weird whitespace here.

go.mod
```diff
@@ -8,28 +8,33 @@ require (
 	github.com/mitchellh/cli v1.1.5
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/sirupsen/logrus v1.9.3
+	golang.org/x/oauth2 v0.15.0
 	gopkg.in/yaml.v2 v2.4.0
 )
 
 require (
 	github.com/Masterminds/goutils v1.1.1 // indirect
-	github.com/Masterminds/semver/v3 v3.2.1 // indirect
```
ChuckCrawford, line 16: It looks like this PR is currently reverting many of the version upgrades from https://github.com/PagerDuty/go-pagerduty/pull/504. If we can get that sorted out then we can get this merged...

ChuckCrawford (approved):
Very exciting!

mishranant:
Hi folks, when will this be released? I see latest release is v1.8.0 and there is no mention of OAuthScope in that.

PR #503, opened 2023-12-13
[CSGI-2365] Add client User Agent overwrite capability

imjaroiswebdev:
## Test cases introduced...

```sh
$ go test -count=1 -v -run TestClient_UserAgent
=== RUN   TestClient_UserAgentDefault
--- PASS: TestClient_UserAgentDefault (0.00s)
=== RUN   TestClient_UserAgentOverwrite
--- PASS: TestClient_UserAgentOverwrite (0.00s)
PASS
ok      github.com/PagerDuty/go-pagerduty       0.667s

```

### DIFF
--- a/client.go
+++ b/client.go
@@ -285,6 +285,8 @@ type Client struct {
 	// PagerDuty API. You can use either *http.Client here, or your own
 	// implementation.
 	HTTPClient HTTPClient
+
+	userAgent string
 }
 
 // NewClient creates an API client using an account/user API token
@@ -319,6 +321,14 @@ type ClientOptions func(*Client)
 func WithAPIEndpoint(endpoint string) ClientOptions {
 	return func(c *Client) {
 		c.apiEndpoint = endpoint
+	}
+}
+
+// WithTerraformProvider configures the client to be used as the PagerDuty
+// Terraform provider
+func WithTerraformProvider(version string) ClientOptions {
+	return func(c *Client) {
+		c.userAgent = fmt.Sprintf("(%s %s) Terraform/%s", runtime.GOOS, runtime.GOARCH, version)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -285,6 +285,9 @@ type Client struct {
 	// PagerDuty API. You can use either *http.Client here, or your own
 	// implementation.
 	HTTPClient HTTPClient
+
+	// UserAgent allows to overwrite Client User Agent value
+	UserAgent string
```
ChuckCrawford, line 290: 1) Could we please use the `ClientOptions` / `WithUserAgent` pattern for setting this value? (See `WithAPIEndpoint(endpoint string) ClientOptions` for an example of this.

2) Having the User-Agent based usage data is pretty valuable. What do you think about limiting the ability for clients to customize this? This might look something like:
```go
func WithTerraformProvider(version string) ClientOptions {
	return func(c *Client) {
		c.userAgent = fmt.Sprintf("Terraform/%s", version)
	}
}
```

client.go
```diff
@@ -498,7 +501,11 @@ func (c *Client) prepRequest(req *http.Request, authRequired bool, headers map[s
 		}
 	}
 
-	req.Header.Set("User-Agent", userAgentHeader)
+	userAgent := userAgentHeader
```
ChuckCrawford, line 504: nit: A little indirect to set the variable and then override it. Consider:
```go
	var userAgent string
	if c.UserAgent != "" {
		userAgent = c.UserAgent
	} else {
		userAgent = userAgentHeader
	}
```

ChuckCrawford (changes requested):
See the inline comments.

client.go
imjaroiswebdev: fair enough 👍🏽

client.go
```diff
@@ -285,6 +285,9 @@ type Client struct {
 	// PagerDuty API. You can use either *http.Client here, or your own
 	// implementation.
 	HTTPClient HTTPClient
+
+	// userAgent allows to overwrite Client User Agent value
```
ChuckCrawford, line 289: nit: maybe remove this comment since we no longer directly allow customization.

client.go
```diff
@@ -322,6 +325,14 @@ func WithAPIEndpoint(endpoint string) ClientOptions {
 	}
 }
 
+// WithTerraformProvider allows to pass in the Terraform version utilized by
```
ChuckCrawford, line 328: nit: maybe we could update this comment to more directly describe what we are doing here?  Something like: "WithTerraformProvider configures the client to be used as the PagerDuty Terraform provider"

ChuckCrawford (approved):
Consider updating the few remaining nits but i'll go ahead and approve to keep this unblocked.

ChuckCrawford (approved):

PR #508, opened 2024-01-09
Prepares for 1.9.0 development.

ChuckCrawford:
Updates client version for 1.9.0 development.

### DIFF
--- a/client.go
+++ b/client.go
@@ -25,7 +25,7 @@ import (
 )
 
 // Version is current version of this client.
-const Version = "1.8.0"
+const Version = "1.9.0-alpha"
 
 const (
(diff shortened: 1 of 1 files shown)


ChuckCrawford:
Follow-up of: https://github.com/PagerDuty/go-pagerduty/pull/507

gaberduty (approved):
This won't be an issue if, e.g., we need to release a security fix that doesn't change any functionality, which would normally be `1.8.1`, right?

ChuckCrawford:
> This won't be an issue if, e.g., we need to release a security fix that doesn't change any functionality, which would normally be `1.8.1`, right?

Correct. We can always cut a branch from the `v1.8.0` commit and work from there if necessary.

PR #521, opened 2024-03-15
[CSGI-3093] Add rate limit headers handling

imjaroiswebdev:
Implements the handling of PagerDuty REST API [ratelimit headers](https://developer.pagerduty.com/docs/72d3b724589e3-rest-api-rate-limits), through an exponential backoff, plus Jitter algorithm. Described [here](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).

This will make part of the solution to https://github.com/PagerDuty/terraform-provider-pagerduty/issues/824, once this client get updated into PagerDuty TF provider dependencies.

### DIFF
--- a/client.go
+++ b/client.go
@@ -14,10 +14,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
+	"math/rand"
 	"net"
 	"net/http"
 	"path"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -33,6 +36,7 @@ const (
 	apiEndpoint         = "https://api.pagerduty.com"
 	identityEndpoint    = "https://identity.pagerduty.com"
 	v2EventsAPIEndpoint = "https://events.pagerduty.com"
+	jitterPercent       = 0.3
 )
 
 // The type of authentication to use with the API client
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -615,11 +619,55 @@ func (c *Client) doWithEndpoint(ctx context.Context, endpoint, method, path stri
...
+	resp, err := c.HTTPClient.Do(req)
+
+	if err != nil || resp.StatusCode != http.StatusTooManyRequests {
+		return resp, err
+	}
+
+	// Rate limited, wait and retry
+	var delay time.Duration
+	rateLimitReset := resp.Header.Get("ratelimit-reset")
+	resetTime, err := strconv.ParseInt(rateLimitReset, 10, 64)
```
cjgajard, line 640: I think you should rename `err` in this line to something else. Currently in the case that `rateLimitReset` is an invalid integer, the user will lose the `url.Error` returned by `c.HTTPClient.Do(req)` which provides more useful information and is more common to be handled explicitly, which also adds the possibility of breaking a client if he assumed Do always returns an `url.Error`.
```
resp, err := client.Something()
if err != nil {
    if urlErr := err.(url.Error); urlErr.Timeout() { // This will panic
        // handles timeout...
    }
}

imjaroiswebdev: Oh Thank you! You are right. Good catch @cjgajard

ChuckCrawford (commented):
There are a few issues with this PR as it currently stands. Some of the bigger ones include:
* The fact that it will not work for any request that contains a `Body`.
* The [slight?] possibility that it could retry forever.
* The fact that it is not configurable and enabled by default.

I put together an alternate implementation in this PR: https://github.com/PagerDuty/go-pagerduty/pull/522 and would be interested to know whether it works for your use-cases.

ChuckCrawford:
Now that https://github.com/PagerDuty/go-pagerduty/pull/522 is merged we should be able to close this @imjaroiswebdev

PR #522, opened 2024-03-20
An initial implementation of client retries.

ChuckCrawford:
# Overview

Adds the ability to configure the client to retry certain types of failures.

## Changes
When configured for retries: 
* The client will retry any request that fails with a network error, 5xx, or 429 status code.
* The client will back off before retrying a request:
  * Server-provided indicators of how long to wait will be used when available.
  * Otherwise we use a truncated binary exponential backoff with jitter.

## Other Notes
Note that this functionality is not supported in the following API functions at this time:
* `pagerduty.CreateEvent`
* `pagerduty.CreateEventWithHTTPClient`
* `client.Do`

This likely be optional functionality for the current 1.9 release. Please give us your feedback.

### DIFF
--- a/client.go
+++ b/client.go
@@ -14,10 +14,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"net/http"
 	"path"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -273,6 +275,16 @@ type HTTPClient interface {
 // Keep this unexported so consumers of the package can't make changes to it.
 var defaultHTTPClient HTTPClient = newDefaultHTTPClient()
 
+type retryPolicy struct {
+	MaxDelay   time.Duration
+	MaxRetries int
+}
+
+var defaultRetryPolicy = retryPolicy{
+	MaxDelay:   20 * time.Second,
+	MaxRetries: 0,
+}
+
 // Client wraps http client
 type Client struct {
 	debugFlag    *uint64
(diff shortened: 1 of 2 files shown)


Sjeanpierre (approved):

client.go
```diff
@@ -273,6 +275,16 @@ type HTTPClient interface {
 // Keep this unexported so consumers of the package can't make changes to it.
 var defaultHTTPClient HTTPClient = newDefaultHTTPClient()
 
+type retryPolicy struct {
+	MaxDelay   time.Duration
+	MaxRetries int
+}
+
+var defaultRetryPolicy = retryPolicy{
```
tomhart-msc, line 283: I think you should have a space between `retryPolicy` and `{`? Have you run `go fmt`?

client.go
```diff
@@ -580,9 +605,61 @@ func dupeRequest(r *http.Request) (*http.Request, error) {
...
+			return nil, err
+		}
+	}
+
+	// Handle final results
+	if respErr != nil {
+		return nil, respErr
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```
tomhart-msc, line 650: Do we want to return 3xx as an error?

tomhart-msc (approved):
LGTM, minor comments.

client.go
ChuckCrawford: Yes. I have. In Go the bracket comes right after the struct name: 

```go
package main

import "fmt"

type Vertex struct {
	X int
	Y int
}

func main() {
	fmt.Println(Vertex{1, 2})
}
```
[Source](https://go.dev/tour/moretypes/2)

PR #541, opened 2024-10-04
fix: Support the From header when snoozing an incident

adamdecaf:
The API docs list From as a required header.

https://developer.pagerduty.com/api-reference/fc871396fb23f-snooze-an-incident#request-headers

### DIFF
--- a/incident.go
+++ b/incident.go
@@ -540,21 +540,29 @@ func (c *Client) CreateIncidentNote(id string, note IncidentNote) error {
 	return err
 }
 
+type SnoozeIncidentOptions struct {
+	From     string
+	Duration uint
+}
+
 // SnoozeIncidentWithResponse sets an incident to not alert for a specified
 // period of time.
 //
 // Deprecated: Use SnoozeIncidentWithContext instead.
-func (c *Client) SnoozeIncidentWithResponse(id string, duration uint) (*Incident, error) {
-	return c.SnoozeIncidentWithContext(context.Background(), id, duration)
+func (c *Client) SnoozeIncidentWithResponse(id string, o SnoozeIncidentOptions) (*Incident, error) {
+	return c.SnoozeIncidentWithContext(context.Background(), id, o)
 }
 
 // SnoozeIncidentWithContext sets an incident to not alert for a specified period of time.
-func (c *Client) SnoozeIncidentWithContext(ctx context.Context, id string, duration uint) (*Incident, error) {
-	d := map[string]uint{
-		"duration": duration,
+func (c *Client) SnoozeIncidentWithContext(ctx context.Context, id string, o SnoozeIncidentOptions) (*Incident, error) {
+	headers := map[string]string{
+		"From": o.From,
+	}
+	body := map[string]uint{
+		"duration": o.Duration,
 	}
 
-	resp, err := c.post(ctx, "/incidents/"+id+"/snooze", d, nil)
+	resp, err := c.post(ctx, "/incidents/"+id+"/snooze", body, headers)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


adamdecaf:
This is a breaking change but the snooze calls don't work currently, so they're already broken.

adamdecaf:
cc @ChuckCrawford  @imjaroiswebdev

incident.go
```diff
@@ -544,17 +544,20 @@ func (c *Client) CreateIncidentNote(id string, note IncidentNote) error {
...
+func (c *Client) SnoozeIncidentWithResponse(id, from string, duration uint) (*Incident, error) {
+	return c.SnoozeIncidentWithContext(context.Background(), id, from, duration)
 }
 
 // SnoozeIncidentWithContext sets an incident to not alert for a specified period of time.
-func (c *Client) SnoozeIncidentWithContext(ctx context.Context, id string, duration uint) (*Incident, error) {
-	d := map[string]uint{
+func (c *Client) SnoozeIncidentWithContext(ctx context.Context, id, from string, duration uint) (*Incident, error) {
+	headers := map[string]string{
+		"from": from,
```
ChezCrawford, line 554: nit: use upper-case header names to be consistent with the other methods and the documentation.
```suggestion
		"From": from,
```

incident.go
```diff
@@ -570,10 +573,13 @@ func (c *Client) SnoozeIncidentWithContext(ctx context.Context, id string, durat
 // SnoozeIncident sets an incident to not alert for a specified period of time.
 //
 // Deprecated: Use SnoozeIncidentWithContext instead.
-func (c *Client) SnoozeIncident(id string, duration uint) error {
+func (c *Client) SnoozeIncident(id, from string, duration uint) error {
+	headers := make(map[string]string)
+	headers["from"] = from
```
ChezCrawford, line 578: nit: same as above.
```suggestion
	headers["From"] = from
```

incident_test.go
```diff
@@ -670,7 +670,7 @@ func TestIncident_SnoozeIncident(t *testing.T) {
 	var duration uint = 3600
 	id := "1"
 
-	err := client.SnoozeIncident(id, duration)
```
ChezCrawford, line 673: It would be nice if we actually verified the `From` header was being sent here in the test. You can see https://github.com/PagerDuty/go-pagerduty/blob/0ef8f340dd3cab97a37f8149760b64a45510c036/incident_test.go#L958-L1007 as an example.

ChezCrawford (changes requested):
If you don't mind addressing the tiny comments I would be happy to get this merged.

incident.go
adamdecaf: Fixed

incident.go
adamdecaf: Fixed

incident_test.go
adamdecaf: Fixed

adamdecaf:
@ChezCrawford thanks - should be g2g now

ChezCrawford:
Ahhh. As I am coming out of my holiday-vacation-fog I am realizing that this is actually a breaking change for some users. The `From` header is only required when using an Account API Key or an App OAuth Token. Clients authenticating with User API Keys are currently able to use this function without issue.

If we are going to make a breaking change here I am tempted to ask that we do this the _right way_ by adding a `SnoozeIncidentOptions` struct with `From` and `Duration` fields. Since the other "write incident" functions currently work the way you have this implemented we can leave this as-is.

adamdecaf:
I can make that change to use a struct. Yes, it's breaking but for me Snooze wasn't even working without this.

ChezCrawford:
> Yes, it's breaking but for me Snooze wasn't even working without this.

You are likely using one of the two types of tokens I mentioned above. I did verify that it works correctly with User API Keys right now.

>  I can make that change to use a struct.

I think that is the best move here. If we are going to make a breaking change let's change it to the "correct" thing that will work properly for everyone.

ChuckCrawford (approved):
Note that this PR makes a breaking change in order to add support for the snooze incident API with Account API Keys and App Tokens.

PR #544, opened 2024-10-29
Fix Jira Cloud accounts mapping Update

cjgajard:
Fixes the Update call for Jira Cloud's account mapping Rule, because the API expect the body of the PUT method to be different to the body of the Create call (e.g. must not have `.config.service`) but currently they are equal.

## [Create](https://developer.pagerduty.com/api-reference/85dc30ba966a6-create-a-rule):
image

## [Update](https://developer.pagerduty.com/api-reference/bce6c10fa07d7-update-a-rule):
image

### DIFF
--- a/jira_cloud_accounts_mapping.go
+++ b/jira_cloud_accounts_mapping.go
@@ -31,8 +31,8 @@ type PagerDutyAccount struct {
 	Subdomain string `json:"subdomain"`
 }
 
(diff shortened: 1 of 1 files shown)


imjaroiswebdev (approved):
LGTM 👍

ChuckCrawford (approved):

PR #545, opened 2024-11-11
fix: return empty list if there are no alertgroupingsettings

alxndr13:
fixes https://github.com/PagerDuty/terraform-provider-pagerduty/issues/944


imjaroiswebdev (approved):
Thank you very much for fixing this @alxndr13 🏆

PR #548, opened 2025-01-23
Merge v1.8.0 to get AssignTagWithContext

tompreston:
Merge v1.8.0 to get [AssignTagWithContext](https://github.com/PagerDuty/go-pagerduty/blob/v1.8.0/tag.go#L155). This will let us [assign tags](https://developer.pagerduty.com/api-reference/7ff9870a1fb8d-assign-tags) which will let us ignore Escalation Policies from Incident IO.

We've forked this for two reasons:
1. We added contact.go, for `ListContactMethods` and `GetContactMethod`
2. In incident.go we added `UpdateIncident` to update the incident severity

We should consider upstreaming these changes.


tompreston:
Wrong repo, please ignore this.

PR #557, opened 2025-06-20
feat: add incident analytics by escalation policy

walbertus:
This MR is to add capabilities defined in this page
https://developer.pagerduty.com/api-reference/8b369ae2eff5e-get-aggregated-escalation-policy-data

1. Add missing filter (escalation_policy_ids, min ack and min manual escalations)
2. Add missing response (escalation policy ID and escalation policy Name)
3. Add missing method GetAggregatedEscalationPolicyData

One thing that I'm not sure, please confirm:
in the docs, it's using min_ackowledgements, but this is a typo, it should be min_acknowledgements. If this block merging than I'll remove this changes from my MR.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -52,14 +52,18 @@ type AnalyticsRawIncidentsResponse struct {
 // AnalyticsFilter represents the set of filters as part of the request to PagerDuty when
 // requesting analytics.
 type AnalyticsFilter struct {
-	CreatedAtStart string   `json:"created_at_start,omitempty"`
-	CreatedAtEnd   string   `json:"created_at_end,omitempty"`
-	Urgency        string   `json:"urgency,omitempty"`
-	Major          bool     `json:"major,omitempty"`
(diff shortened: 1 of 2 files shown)


walbertus:
@cjgajard please check on this if you're available, thank you

walbertus:
I'm checking with the API for min_acknowledgements vs min_ackowledgements, it seems the documentation is typo. The API is using min_acknowledgements.

cjgajard (changes requested):
Thank you for your contribution 🎉 

AnalyticsFilter missing pd_advance_used
image

AnalyticsRequest missing aggregate_unit (it should be `*string` without `,omitempty` to allow sending a `null`)
image

walbertus:
@cjgajard I've updated the pg_advanced_use

for the aggregate_unit, I'd rather not update that to *string for two reasons:
1. It is outside of the scope of this MR
2. It introduces breaking changes to the API user

It currently using string here https://github.com/PagerDuty/go-pagerduty/blob/main/analytics.go#L15

PR #558, opened 2025-07-15
Fix Update for ServiceCustomField

cjgajard:
Could not update service custom field P123456: HTTP response failed with status code 400, message: Invalid Input Provided. (code: 2001): field.name: doesn't match any valid key

and

Provider produced inconsistent result after apply


imjaroiswebdev (approved):
👍🏽

PR #560, opened 2025-08-11
Internal Refactoring: Enhanced response parsing for different API endpoint formats in enablements

imjaroiswebdev:
- Refactored enablements API response parsing to properly handle different endpoint formats between services and event
   orchestrations
- Updated API methods to return warnings in response structures instead of logging them internally
- Improved error handling and response decoding logic for better API compatibility


cjgajard (approved):
🎉

PR #562, opened 2025-09-09
Add device_type to Contact Method

cjgajard:
Required for creation of PushContactMethods

`Error: POST API call to https://api.pagerduty.com/users/P123456/contact_methods failed 400 Bad Request. Code: 2001, Errors: [Device type can't be blank Device type is not included in the list]`


imjaroiswebdev (approved):
Good! Ty! 💪

PR #563, opened 2025-09-25
Fix version flag to use git tag

phumpal:
May be intentional but `pd --version` will always return `0.0.0.` as `version` is a constant not a var.


cjgajard (approved):
LGTM!

